PR #4, opened 2016-10-19
Add support for 3D Secure

lkorth:


Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
```diff
@@ -161,6 +172,19 @@ public DropInRequest disableVenmo() {
...
 
     /**
+     * If 3D Secure has been enabled in the control panel and an amount is specified in
+     * {@link DropInRequest#amount(String)}, Drop-In will request a 3D Secure verification for
+     * any new cards added by the user.
+     *
+     * @param requestThreeDSecure {@code true} to request a 3D Secure verification as part of Drop-In,
+     * {@code false} to not request a 3D Secure verification. Defaults to {@code false}.
+     */
+    public DropInRequest requestThreeDSecureVerification(boolean requestThreeDSecure) {
```
sdcoffey: `DropinRequest` is more or less a one-time use object, it doesn't really seem like someone would disable 3DS after enabling it. Would it make more sense for this method to take a `String amount` instead of a boolean?

sdcoffey: Also, why did we even have an `mAmount` member if there were no getters or setters previously?

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
```diff
@@ -63,6 +64,16 @@ public DropInRequest tokenizationKey(String tokenizationKey) {
     }
 
     /**
+     * This method is optional. Amount is only used for 3D Secure verifications.
+     *
+     * @param amount Amount of the transaction.
```
sdcoffey: Does the optional tag here indicate that it's optional for drop-in, or optional for 3DS? My understanding was that an amount was required for 3DS.

lkorth: It's optional in general for the `DropInRequest`.

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
lkorth: It is a one time use object, however to stay more in line with the other methods for disabling and enabling things I went with a boolean here. `mAmount` was a holdover from the previous version of Drop-in that was not completely removed in the new version (getters and setters were removed, but the variable was missed). Amount was previously used for display in the UI and could still be used in the future for UI or other purposes.

PR #17, opened 2017-01-06
Filter Android Pay payment methods from the list of vaulted payment methods

lkorth:
The [docs to not recommend vaulting Android Pay payment methods](https://developers.braintreepayments.com/guides/android-pay/server-side), but when they are vaulted, they should not show up in the list of vaulted payment methods in Drop-in.


Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -260,11 +261,19 @@ public void run() {
     }
 
     @Override
-    public void onPaymentMethodNoncesUpdated(final List<PaymentMethodNonce> paymentMethodNonces) {
-        if (paymentMethodNonces.size() > 0) {
+    public void onPaymentMethodNoncesUpdated(List<PaymentMethodNonce> paymentMethodNonces) {
+        List<PaymentMethodNonce> filteredPaymentMethodNonces = new ArrayList<>(paymentMethodNonces);
```
sdcoffey: Dang, would be nice if we could use the shiny new new `List.removeIf()` here.

lkorth: 😞

quinnjn:
LGTM! 👍  🎆

PR #29, opened 2017-03-24
Respect DropInRequest payment method overrides

### LINKED ISSUE #27: No way to disable PayPal using DropInRequest
### General information

* SDK/Library version: 3.0.6
* Environment: all
* Android Version and Device: doesn't matter

### Issue description

Hi, I've tried using the `DropInRequest.disablePaypal()` method but it does nothing, and looking through the code of the `SupportedPaymentMethodsAdapter`, it's just checking the `Configuration` to see if Paypal is enabled, but the explicit flag passed in `DropInRequest` is not used.

lkorth:
Fixes #27

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/SupportedPaymentMethodsAdapter.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/SupportedPaymentMethodsAdapter.java
@@ -9,6 +9,7 @@ import android.widget.BaseAdapter;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import com.braintreepayments.api.dropin.DropInRequest;
 import com.braintreepayments.api.dropin.R;
 import com.braintreepayments.api.dropin.utils.PaymentMethodType;
 import com.braintreepayments.api.models.Configuration;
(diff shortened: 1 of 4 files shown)


Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/SupportedPaymentMethodsAdapter.java
```diff
@@ -23,19 +24,20 @@
     private ArrayList<PaymentMethodType> mAvailablePaymentMethods;
     private PaymentMethodSelectedListener mPaymentMethodSelectedListener;
 
-    public SupportedPaymentMethodsAdapter(Context context, Configuration configuration,
-                                          boolean androidPayEnabled, boolean unionpaySupported,
+    public SupportedPaymentMethodsAdapter(Context context,
```
quinnjn: This is technically a breaking change, even though I don't think a of folks would be using this adapter.

sdcoffey:
Agree with Quinn about the breaking change, but I doubt that anyone is using that. I'm willing to take the risk 🍏

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/SupportedPaymentMethodsAdapter.java
lkorth: This is internal to Drop-in and undocumented.

Issue #41, opened 2017-06-29
DropInActivity remains in progress state upon leaving ThreeDSecureWebViewActivity 


consp1racy:
### General information

* SDK/Library version: drop-in:3.0.8, braintree:2.5.2, card-form:3.0.6
* Environment: Any
* Android Version and Device: Any

### Issue description

When leaving ThreeDSecureWebViewActivity using the system back button or action bar up button underlying Drop-In activity is left in progress state.

### Expected behavior

Upon leaving ThreeDSecureWebViewActivity Drop-In activity would hide progress and show the list of recent and availabel payment methods again.

lkorth:
Thanks for the report @consp1racy. I just opened #43 to fix this.

PR #44, opened 2017-08-15
Run 3D Secure verifications on vaulted cards

lkorth:
The second half of fixes for #41. 

This change runs 3D Secure verifications on vaulted cards when 3D Secure is requested and enabled ensuring that all cards returned by Drop-in have run through a 3D Secure verification.

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/BaseActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/BaseActivity.java
@@ -1,0 +1,43 @@
+package com.braintreepayments.api.dropin;
+
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.support.v7.app.AppCompatActivity;
+import android.text.TextUtils;
+
+import com.braintreepayments.api.BraintreeFragment;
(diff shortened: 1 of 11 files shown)


Drop-In/src/main/java/com/braintreepayments/api/dropin/BaseActivity.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        mDropInRequest = getIntent().getParcelableExtra(DropInRequest.EXTRA_CHECKOUT_REQUEST);
+    }
+
+    protected BraintreeFragment getBraintreeFragment() throws InvalidArgumentException {
+        if (TextUtils.isEmpty(mDropInRequest.getAuthorization())) {
+            throw new InvalidArgumentException("A client token or client key must be specified " +
```
quinnjn: Should we still mention `client key`?

lkorth: :man_facepalming: nope, I'll rename this

PR #84, opened 2018-08-15
Delete payment methods

quinnjn:
Allows customers to delete saved payment methods through Drop-in.

delete-payment-methods

TODOs before release
* Translations
* UI confirmation
* Release braintree-android and replace the SNAPSHOT dependency


Demo/src/androidTest/java/com/braintreepayments/demo/test/DropInTest.java
```diff
@@ -216,6 +220,31 @@ public void exitsAfterCancelingAddingAPaymentMethod() {
         onDevice(withText("Add Payment Method")).check(text(equalToIgnoringCase("Add Payment Method")));
     }
 
+    @Test(timeout = 60000)
+    public void deletesPaymentMethod() {
```
jackellenberger: Do we need to do any verification that the customer we choose does not have existing payment methods? Or do we ensure that we clean up after ourselves somewhere else?

quinnjn: I think I can play around with more assumptions to verify that we're in the right state for this test.

Drop-In/src/main/java/com/braintreepayments/api/dropin/VaultManagerActivity.java
```diff
@@ -0,0 +1,143 @@
...
+                        public void onDismiss(DialogInterface dialog) {
+                            if (!positiveSelected.get()) {
+                                mAdapter.cancelSwipeOnPaymentMethodNonce(paymentMethodNonceToDelete);
+                                mBraintreeFragment.sendAnalyticsEvent("manager.delete.confirmation.negative");
+                            }
+                        }
+                    })
+                    .setNegativeButton(R.string.bt_cancel, new DialogInterface.OnClickListener() {
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {
```
jackellenberger: Will this dismiss the dialog?

yagrawl: Yes. `AlertDialog.Builder` will call the `OnDismissListener` on either button click. So within that listener we check if the positive button wasn't selected and cancel the swipe.

Drop-In/src/main/res/layout/bt_drop_in_activity.xml
```diff
@@ -55,17 +55,36 @@
...
+                            android:layout_alignParentRight="true"
+                            android:text="@string/bt_edit"/>
+
+                    </RelativeLayout>
+
+                    <android.support.v7.widget.RecyclerView
+                        android:id="@+id/bt_vaulted_payment_methods"
+                        android:layout_width="match_parent"
+                        android:layout_height="wrap_content"
+                        android:paddingLeft="4dp"
```
yagrawl: Should we be referencing this padding and a few others down below from `@dimen` instead?

Drop-In/src/main/res/values/strings.xml
```diff
@@ -33,4 +33,12 @@
...
     <string name="bt_unionpay_sms_code_invalid">SMS code is invalid</string>
     <string name="bt_unionpay_mobile_number_explanation">Enrollment may be required for this card. An SMS code will be sent to this number if required.</string>
+    <string name="bt_edit">Edit</string>
+    <string name="bt_delete">Delete</string>
+    <string name="bt_cancel">@android:string/cancel</string>
+    <string name="bt_vault_manager_title">Edit Saved</string>
+    <string name="bt_delete_confirmation_title">Delete payment method?</string>
+    <string name="bt_delete_confirmation_description">You will no longer be able to use this payment method in this app.</string>
+    <string name="bt_done">Done</string>
+    <string name="bt_vault_manager_delete_failure">We couldn\'t delete your payment method right now, try again later</string>
```
yagrawl: Missing period at the end of this message.

Drop-In/src/main/java/com/braintreepayments/api/dropin/VaultManagerActivity.java
jackellenberger: 🙏  thanks Yash!

quinnjn:
* Added a loading view while its deleting the payment method
* Returns the current payment methods back to DropInActivity instead of force reloading the payment methods

android-delete

Drop-In/src/main/res/layout/bt_drop_in_activity.xml
```diff
@@ -55,17 +55,36 @@
...
+
+                        <Button
+                            style="@style/Widget.AppCompat.Button.Borderless.Colored"
+                            android:theme="@style/bt_edit_button"
+                            android:id="@+id/bt_vault_edit_button"
+                            android:layout_width="wrap_content"
                             android:layout_height="wrap_content"
-                            android:paddingLeft="4dp"
-                            android:paddingRight="4dp"/>
+                            android:paddingTop="@dimen/bt_choose_existing_payment_label_padding_top"
```
jackellenberger: I think this is throwing off the centering of the text within the touch target which makes the touch ripple look kinda off

Drop-In/src/main/res/values/styles.xml
```diff
@@ -32,6 +32,19 @@
...
     </style>
 
+    <style name="bt_vault_manager_activity_theme" parent="Theme.AppCompat.NoActionBar">
+        <item name="windowNoTitle">true</item>
+        <item name="android:windowContentOverlay">@null</item>
+        <item name="android:windowBackground">@android:color/transparent</item>
+        <item name="android:colorBackgroundCacheHint">@null</item>
+        <item name="android:windowIsTranslucent">true</item>
+        <item name="android:windowTranslucentStatus" tools:targetApi="kitkat">true</item>
+        <item name="android:windowAnimationStyle">@style/bt_fade_activity_animation</item>
```
jackellenberger: was the slide animation too buggy? in the first gif there was some pretty bad clipping, but i think the fade is kinda jarring, unless we can get the bottom sheet to be the same height as the select payment method view

Issue #77, opened 2018-06-26
textColorError - SDK 28


StefanDeBruijn:
The use of `textColorError` in the style `bt_add_card_activity_theme` blocks users from building with SDK and support library 28

quinnjn:
Confirmed, thanks for the report!

We are working on upgrading Braintree Android in preparation for Android P: https://github.com/braintree/braintree_android/issues/210

After we're done upgrading Braintree Android, Drop-in will also get an Android P update.

StefanDeBruijn:
Alright, great to hear. Thanks

quinnjn:
We can keep this issue open until it is fixed

Klingefjord:
@dschuermann Could you clarify this workaround? Doesn't seem to work for me - what level do you declare foo in in attrs.xml and styles.xml?

quinnjn:
I merged in #93 which solves the problem and published a [snapshot version](https://oss.sonatype.org/content/repositories/snapshots/com/braintreepayments/api/drop-in/3.6.2-SNAPSHOT/).

Please try it out and let us know if it resolves the issue!

Klingefjord:
@quinnjin resolved the issue, thanks! 
Any estimate when this will be out in a proper release?

quinnjn:
Glad it worked! We're working on tying up one last feature before the next release. It will be out by EOD tomorrow.

quinnjn:
[3.7.0](https://github.com/braintree/braintree-android-drop-in/releases/tag/3.7.0) is now available.

```
implementation 'com.braintreepayments.api:drop-in:3.7.0'
```

git-yourdevexpert:
when i add it in build.gradle file it gives error 

  implementation 'com.braintreepayments.api:drop-in:3.7.0'

error: style attribute 'attr/colorError (aka com.kudos.diplomakids:attr/colorError)' not found.
  error: failed linking references.

Manaspaul623:
> This fix is now released and available in [3.7.0](https://github.com/braintree/braintree-android-drop-in/releases/tag/3.7.0).
> 
> ```
> implementation 'com.braintreepayments.api:drop-in:3.7.0'
> ```

Worked for me

jjmcgarry90:
Still having this issue

exaland:
> This fix is now released and available in [3.7.0](https://github.com/braintree/braintree-android-drop-in/releases/tag/3.7.0).
> 
> ```
[...]

PR #93, opened 2018-09-13
[Blocked on card-form release] Update SDK to 28

quinnjn:
* Bumps SDK to 28
* Replaces `compile` dependency keyword
* Removes mailable log
* Moves `support-v4` resolution strategy to Drop-in


quinnjn:
I'm removing the block since I would like to get a snapshot out for the folks in issue #77. 

We should hold on releasing Drop-in until android-card-form@3.5.0 gets released.

PR #130, opened 2019-05-13
Add analytic events for vaulted card

scannillo:
We previously collected analytic events for tokenization on new card creation only. This PR adds an analytic event for choosing an already vaulted card.

Events:
- Start event "vaulted-card.appear": when all vaulted cards are shown
- Success event "vaulted-card.select": when a vaulted card is chosen

Notes:
- Will squash after review and before merge.

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
@@ -13,6 +13,7 @@ import android.widget.ListView;
 import android.widget.TextView;
 import android.widget.ViewSwitcher;
 
+import com.braintreepayments.api.Card;
 import com.braintreepayments.api.DataCollector;
 import com.braintreepayments.api.GooglePayment;
 import com.braintreepayments.api.PayPal;
@@ -274,12 +275,28 @@ public class DropInActivity extends BaseActivity implements ConfigurationListene
         if (paymentMethodNonces.size() > 0) {
             mSupportedPaymentMethodsHeader.setText(R.string.bt_other);
             mVaultedPaymentMethodsContainer.setVisibility(View.VISIBLE);
-            mVaultedPaymentMethodsView.setAdapter(new VaultedPaymentMethodsAdapter(this,
-                    paymentMethodNonces));
+            mVaultedPaymentMethodsView.setAdapter(new VaultedPaymentMethodsAdapter(new PaymentMethodNonceCreatedListener() {
+                @Override
+                public void onPaymentMethodNonceCreated(PaymentMethodNonce paymentMethodNonce) {
+                    if (paymentMethodNonce instanceof CardNonce) {
+                        mBraintreeFragment.sendAnalyticsEvent("vaulted-card.select");
+                    }
+
+                    DropInActivity.this.onPaymentMethodNonceCreated(paymentMethodNonce);
+                }
+            }, paymentMethodNonces));
 
             if (mDropInRequest.isVaultManagerEnabled()) {
                 mVaultManagerButton.setVisibility(View.VISIBLE);
             }
+
+            for (PaymentMethodNonce nonce : paymentMethodNonces) {
+                if (nonce instanceof CardNonce) {
+                    mBraintreeFragment.sendAnalyticsEvent("vaulted-card.appear");
+                    break;
+                }
+            }
+
         } else {
             mSupportedPaymentMethodsHeader.setText(R.string.bt_select_payment_method);
             mVaultedPaymentMethodsContainer.setVisibility(View.GONE);
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/VaultedPaymentMethodsAdapter.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/VaultedPaymentMethodsAdapter.java
@@ -23,7 +23,7 @@ public class VaultedPaymentMethodsAdapter extends RecyclerView.Adapter<VaultedPa
     private List<PaymentMethodNonce> mPaymentMethodNonces;
 
     public VaultedPaymentMethodsAdapter(PaymentMethodNonceCreatedListener listener,
-            List<PaymentMethodNonce> paymentMethodNonces) {
+                                        List<PaymentMethodNonce> paymentMethodNonces) {
         mSelectedListener = listener;
         mPaymentMethodNonces = paymentMethodNonces;
     }
--- a/Drop-In/src/test/java/com/braintreepayments/api/dropin/DropInActivityUnitTest.java
+++ b/Drop-In/src/test/java/com/braintreepayments/api/dropin/DropInActivityUnitTest.java
@@ -63,6 +63,7 @@ import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Matchers.matches;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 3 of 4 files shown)


Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/VaultedPaymentMethodsAdapter.java
```diff
@@ -21,11 +22,14 @@
 
     private PaymentMethodNonceCreatedListener mSelectedListener;
     private List<PaymentMethodNonce> mPaymentMethodNonces;
+    private BraintreeFragment mBraintreeFragment;
```
quinnjn: Ideally views/adapters shouldn't know about any other objects other than the ones they are presenting.

I don't have a super straight forward suggestion for you guys. You could add a listener that sends analytic events and chains the PaymentMethodNonceCreatedListeners. Here is some psuedo code for this: 

```
// Add field in DropInActivity.java
private final PaymentMethodNonceCreatedListener vaultedPaymentMethodSelected = (paymentMethodNonce) -> {
  mBraintreeFragment.sendAnalyticsEvent("vaulted-card.selected")

  DropInActivity.this.onPaymentMethodNonceCreated(paymentMethodNonce)
};

...

mVaultedPaymentMethodsView.setAdapter(vaultedPaymentMethodSelected, ...)
```

scannillo: We found another workaround. We instead including this analytic event on the already existing `onPaymentMethodNonceCreated` function. Check it out [here](https://github.com/braintree/braintree-android-drop-in/pull/130/commits/54272e039a48f921b7bbc89a1a3bd93b4cff20f8#diff-4253f3b4ed47a06cee52c9f47fc90b04R203). 

It fires as expected.

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -275,13 +277,19 @@ public void onPaymentMethodNoncesUpdated(List<PaymentMethodNonce> paymentMethodN
...
-                    paymentMethodNonces, mBraintreeFragment));
+                    paymentMethodNonces));
 
             if (mDropInRequest.isVaultManagerEnabled()) {
                 mVaultManagerButton.setVisibility(View.VISIBLE);
             }
 
-            mBraintreeFragment.sendAnalyticsEvent("vaulted-card.appear");
+            for (PaymentMethodNonce nonce : paymentMethodNonces) {
+                if (nonce.getClass() == CardNonce.class) {
```
quinnjn: Can be replaced with `if (nonce instanceof CardNonce) {`

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -200,6 +200,8 @@ public void onAnimationFinished() {
 
     @Override
     public void onPaymentMethodNonceCreated(final PaymentMethodNonce paymentMethodNonce) {
+        mBraintreeFragment.sendAnalyticsEvent("vaulted-card.select");
```
quinnjn: `onPaymentMethodNonceCreated` is called on any `BraintreeFragment` tokenization result, not just vaulted cards. 

Might be worth adding a test that given a non-vaulted tokenization this analytic event is not called

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -280,6 +282,14 @@ public void onPaymentMethodNoncesUpdated(List<PaymentMethodNonce> paymentMethodN
             if (mDropInRequest.isVaultManagerEnabled()) {
                 mVaultManagerButton.setVisibility(View.VISIBLE);
             }
+
+            for (PaymentMethodNonce nonce : paymentMethodNonces) {
+                if (nonce.getClass() == CardNonce.class) {
+                    mBraintreeFragment.sendAnalyticsEvent("vaulted-card.appear");
+                    break;
```
quinnjn: There may be more than 1 vaulted card

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -280,6 +282,14 @@ public void onPaymentMethodNoncesUpdated(List<PaymentMethodNonce> paymentMethodN
             if (mDropInRequest.isVaultManagerEnabled()) {
                 mVaultManagerButton.setVisibility(View.VISIBLE);
             }
+
+            for (PaymentMethodNonce nonce : paymentMethodNonces) {
+                if (nonce.getClass() == CardNonce.class) {
+                    mBraintreeFragment.sendAnalyticsEvent("vaulted-card.appear");
```
quinnjn: Technically at this point we know vaulted cards exist and could be shown, but we haven't drawn a view for it. I think this might be OK though.

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
scannillo: We want to only send this analytic event once. One time for when vaulted card (or multiple) appear on the screen.

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
scannillo: Gotcha! We will add logic here so that this "vaulted-card.select" is only called when the nonce is of a Card.

That should resolve this issue, correct?

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
quinnjn: Ok that makes sense!

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
quinnjn: Can you try make a PayPal nonce (non-vaulted) and see if it goes in this method?

quinnjn: Its an odd callback since it doesn't seem to get a callback on nonce tokenization without vaulting. I think if that is the case and we can prove that with a test I'm happy with it!

scannillo: Good call! Added tests around this here.

https://github.com/braintree/braintree-android-drop-in/pull/130/commits/2f568923786a07c584f25c105076da318256d716

Drop-In/src/test/java/com/braintreepayments/api/dropin/DropInActivityUnitTest.java
```diff
@@ -566,6 +566,22 @@ public void onPaymentMethodNoncesUpdated_withNoCards_doesNotSendAnalyticEvent()
         verify(mActivity.braintreeFragment, never()).sendAnalyticsEvent("vaulted-card.appear");
     }
 
+    @Test
+    public void onPaymentMethodNoncesUpdated_withPayPal_doesNotSendCardAnalyticEvent() throws JSONException {
```
quinnjn: I think this needs to be changed to `onActivityResult_withPayPal_doesNotSendCardAnalyticEvent`.

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
scannillo: Resolved in [this commit](https://github.com/braintree/braintree-android-drop-in/pull/130/commits/d5d48da09ca22f9b928193056ebd66b38328b50a).

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
scannillo: Resolved in [this commit](https://github.com/braintree/braintree-android-drop-in/pull/130/commits/d5d48da09ca22f9b928193056ebd66b38328b50a).

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/VaultedPaymentMethodsAdapter.java
scannillo: Resolved in [this commit](https://github.com/braintree/braintree-android-drop-in/pull/130/commits/d5d48da09ca22f9b928193056ebd66b38328b50a).

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
scannillo: Yup. This is assuming nothing goes wrong with rendering the vaulted payments after we populate them.

Would you rather we implement another listener for use inside the vaulted payment methods recycler view? Likely here in the [on bind](https://github.com/braintree/braintree-android-drop-in/blob/b92db7ae2b98fd5f3afa3c716e150453167d2074/Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/VaultedPaymentMethodsAdapter.java#L38-L57).

quinnjn:
I mistakenly suggested `captures/` but its supposed to be `captures` to ignore the whole folder. Pushed a fixup for that: 798ec69

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
quinnjn: Nah, I think this is fine 👍

scannillo: 👍

Issue #134, opened 2019-06-26
BUG: DropInActivity missing super call to onActivityResult


xX7:
# General information

* SDK/Library version: DropIn 4.2.0
* Environment: Sandbox & Production
* Android Version and Device: Google Pixel 2 XL / Android 9 - Official Rom

### Issue description

Referencing #122
PR: #126 

@scannillo @billwerges @quinnjn 

As I happened to encounter the same problem as stated in the issue above, I went on looking through the drop-in source code, to find that the `DropInActivity` is missing the call to `super.onActivityResult(...)`, so the` BraintreeFragments onActivityResult` won't get called. 

How to reproduce that behavior:

1. Code a custom implementation of a simple purchase flow with PayPal and attach the `BraintreeFragment` to this activity. 

2. Override `onActivityResult` **and remove the super call.** 

3. `onPaymentMethodNonceCreated` won't be called anymore.
 
**Edit:**

Looks to me as the same "thing" you tried to fix in order to get the Google Payment flow to work. By adding the super call, that should be fixed as well. 

Also: https://stackoverflow.com/questions/6147884/onactivityresult-is-not-being-called-in-fragment

**Edit2:**

Digging deeper I found a proof, that my suggested change will fix PayPal and the Google Payment flow:

The temporary fix for the Google Payment flow was to listen for `requestCode == 79129` while the activity was started with` BraintreeRequestCodes.GOOGLE_PAYMENT = 13593`. Giving the fact, that the `BraintreeFragment` is the only Fragment attached to the activity the `fragmentIndex` should be **1**.

originalRequestCode = changedRequestCode - (indexOfFragment << 16)
13593 = 79129 - ( 1 << 16)
13593 = 79129 - 65536
13593 = 13593

This points out, that the `onActivityResult` call with` requestCode = 79129` isn't due to a bug in the Google Payment activity, but just the fact that it is altered so it reaches it's correct fragment with a call to `super.onActivityResult`

crookedneighbor:
Would you like to open a pull request?

scannillo:
This issue was fixed with this PR: https://github.com/braintree/braintree-android-drop-in/pull/135

Issue #122, opened 2019-04-16
Paypal - impossible to complete payment through Paypal app


venator85:
### General information

* SDK/Library version: drop-in 4.1.0
* Environment: both sandbox and production
* Android Version and Device: OnePlus 5T, Android 9.0

### Issue description

It is impossible to complete a payment if the Paypal app is installed (and is automatically launched instead of the browser).

Steps to reproduce:
- created an empty android project with androidx, added drop-in 4.1.0, performed the Browser switch setup (app doesn't have underscores in the package name):
```xml
<activity android:name="com.braintreepayments.api.BraintreeBrowserSwitchActivity"
    android:launchMode="singleTask">
    <intent-filter>
        <action android:name="android.intent.action.VIEW" />
        <category android:name="android.intent.category.DEFAULT" />
        <category android:name="android.intent.category.BROWSABLE" />
        <data android:scheme="${applicationId}.braintree" />
    </intent-filter>
</activity>
```
- created super simple activity with a sample client token from the docs page ("Try it now"):
```kotlin
class MainActivity : AppCompatActivity() {
    private val REQUEST_CODE = 2048

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)
        findViewById<View>(R.id.btn).setOnClickListener {
            val dropInRequest = DropInRequest()
                .clientToken("eyJ2ZXJ.......1vIjoib2ZmIn0=")
            startActivityForResult(dropInRequest.getIntent(this), REQUEST_CODE)
        }
    }

    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
        if (requestCode == REQUEST_CODE) {
            if (resultCode == Activity.RESULT_OK) {
                val result = data?.getParcelableExtra<DropInResult>(DropInResult.EXTRA_DROP_IN_RESULT)
                Log.e("result $result", this)
            } else if (resultCode == Activity.RESULT_CANCELED) {
                Log.e("canceled", this)
            } else {
                val error = data?.getSerializableExtra(DropInActivity.EXTRA_ERROR) as Exception
                Log.e("error", error, this)
            }
        } else {
            super.onActivityResult(requestCode, resultCode, data)
        }
    }
}
```
- on clicking the button, paypal is selected and the paypal app (7.6.0, enrolled with biometric access and a working production account) is launched automatically:
[...]

PR #135, opened 2019-07-02
(ISSUE #134) Added super call to onActivityResult in DropInActivity

scannillo:
Fix in response to [Issue #134](https://github.com/braintree/braintree-android-drop-in/issues/134) & [Issue #122](https://github.com/braintree/braintree-android-drop-in/issues/122). 

*Problem:* With the PayPal app installed, the drop in would get stuck on an infinite load after app switch back from PayPal. `onActivityResult` was not invoked. This required merchants to call `           super.onActivityResult(requestCode, resultCode, data)` in their `MainActivity.java`'s `onActivityResult` override method.

*Solution:* Add the `super.onActivityResult(requestCode, resultCode, data)` to the `DropInActivity`'s `onActivityResult` method override so merchant doesn't isn't requires to make this call.

Credits:
@xX7

*Note:* Couldn't think of a meaningful test to add for this super call code addition that doesn't already exist in the tests. Any suggestions appreciated.

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
@@ -69,8 +69,8 @@ public class DropInActivity extends BaseActivity implements ConfigurationListene
      */
     public static final String EXTRA_ERROR = "com.braintreepayments.api.dropin.EXTRA_ERROR";
 
-    private static final int ADD_CARD_REQUEST_CODE = 1;
-    private static final int DELETE_PAYMENT_METHOD_NONCE_CODE = 2;
+    public static final int ADD_CARD_REQUEST_CODE = 1;
+    public static final int DELETE_PAYMENT_METHOD_NONCE_CODE = 2;
     private static final String EXTRA_SHEET_SLIDE_UP_PERFORMED = "com.braintreepayments.api.EXTRA_SHEET_SLIDE_UP_PERFORMED";
     private static final String EXTRA_DEVICE_DATA = "com.braintreepayments.api.EXTRA_DEVICE_DATA";
     static final String EXTRA_PAYMENT_METHOD_NONCES = "com.braintreepayments.api.EXTRA_PAYMENT_METHOD_NONCES";
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
@@ -312,29 +312,7 @@ public class DropInActivity extends BaseActivity implements ConfigurationListene
 
     @Override
     protected void onActivityResult(int requestCode, final int resultCode, Intent data) {
-        /*
-         * TODO this a temporary work around to fix the Google Payment flow.
-         *
-         * BraintreeFragment starts the GooglePaymentActivity for result to tokenize
-         * Google Payment. GooglePaymentActivity#onActivityResult is called from Google Payment
-         * and a result is set, then the activity finishes.
-         *
-         * What should happen is BraintreeFragment#onActivityResult should be called.
-         *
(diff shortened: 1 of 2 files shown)


Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -328,6 +328,8 @@ protected void onActivityResult(int requestCode, final int resultCode, Intent da
          * The temporary fix is to forward the data back to BraintreeFragment when we detect
          * the 79129 requestCode.
          */
+        super.onActivityResult(requestCode, resultCode, data);
+
         if (requestCode == 79129) {
```
quinnjn: Since we call `super.onActivityResult`, can this work around be dropped? 

See [Edit2](https://github.com/braintree/braintree-android-drop-in/issues/134#issue-461217287)

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -328,6 +328,8 @@ protected void onActivityResult(int requestCode, final int resultCode, Intent da
          * The temporary fix is to forward the data back to BraintreeFragment when we detect
          * the 79129 requestCode.
          */
+        super.onActivityResult(requestCode, resultCode, data);
```
quinnjn: The comment above does not seem related to this line

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
scannillo: Gotcha! My initial thought was that the workaround was still needed as a backup. But I appreciate you pointing it out again. Reading through that a second time makes sense as to why this workaround can now be removed with the super call added. 👍

scannillo:
> DropInActivity#onActivityResult handles add card, and delete payment method results from Drop-in activities. 
> This changes the view state around, i.e. mLoadingViewSwitcher.setDisplayedChild(0);
If we're ending up here for BraintreeFragment#startActivityForResult we might have to do some work to return early before we start changing with Drop-in view states.

I understand what you mean about guarding to return early before changing view states. But am not sure what case would trigger that later scenario. Could you specify a bit?

Thanks for the feedback!

xX7:
Thanks for the Credits!

I think what @quinnjn means is, that the code including and below `mLoadingViewSwitcher.setDisplayedChild(0);` "now" always gets called as soon as a activity started by `BraintreeFragment#startActivityForResult` finishes, which could lead to unintended behavior. This could be fixed/adjusted by checking for `DropInActivity#onActivityResult` desired requestCode's first, before doing any view state switching...

scannillo:
> Is there a test we can write for this behavior? Or not really?
I thought that the already existing tests around view state would be enough to verify that these changes are appropriate ([example test here](https://github.com/braintree/braintree-android-drop-in/blob/c73f04e61f13071a1f59a70109ca04eea4bd915a/Drop-In/src/test/java/com/braintreepayments/api/dropin/DropInActivityUnitTest.java#L603-L611)). 

However, I supposed we could add a test that tests the `super` function is called every time we call `onActivityResult`. I honestly am not sure what would be the most efficient thing to test and would love some suggestions/feedback.

scannillo:
Also, given my research on how `setDisplayedChild(0)` works, it displays the payment selection view of the drop-in.

So it makes sense to me to add it to these 3 situations in the code:
1) RESULT_CANCELED for ADD_CARD_REQUEST_CODE: The user closed the add card flow and we need to show the Drop-in selector again
2) RESULT_OK for ADD_CARD_REQUEST_CODE: The user submitted a credit card and we have a nonce
3) RESULT_OK for DELETE_PAYMENT_METHOD_NONCE_CODE: We successfully deleted a saved payment method

I know @Epreuve and @quinnjn had some concern with this change and I wanted to get their expertise on this as soon as they're free.

PR #137, opened 2019-07-22
amex has updated their logo

rtchang:
original image source from:
https://about.americanexpress.com/logo/brand/american-express-blue-box-logo-full

resized to fit the sizes used in the app

article describing the new logo updates
https://www.underconsideration.com/brandnew/archives/new_logo_and_identity_for_american_express_by_pentagram.php

> Before submitting this PR, note that we cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.


scannillo:
Hello @rtchang. Thanks for pointing this out and for opening up a PR. Unfortunately, we would have to wait to merge such changes because we need our iOS, Web, and Android drop-ins to be in parity and use the same assets.

This task isn't high priority for us right now, so we won't be able to merge this in the near future.

PR #141, opened 2019-07-25
Add support for ThreeDSecureRequest and 3DS 2

billwerges:


Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
```diff
@@ -151,6 +154,19 @@ public DropInRequest requestThreeDSecureVerification(boolean requestThreeDSecure
         return this;
     }
 
+    /**
+     * This method is optional.
+     *
+     * @param threeDSecureRequest {@link ThreeDSecureRequest} to specify options and additional for 3D Secure.
```
scannillo: Missing the word "information" after "additional"

Drop-In/src/main/java/com/braintreepayments/api/dropin/AddCardActivity.java
```diff
@@ -283,19 +286,26 @@ protected void createCard() {
...
     @Override
     public void onPaymentMethodNonceCreated(PaymentMethodNonce paymentMethod) {
-        mBraintreeFragment.sendAnalyticsEvent("sdk.exit.success");
-        finish(paymentMethod, null);
+        if (shouldRequestThreeDSecureVerification() && mDropInRequest.getThreeDSecureRequest() == null) {
+            ThreeDSecureRequest threeDSecureRequest = new ThreeDSecureRequest().amount(mDropInRequest.getAmount());
+            mDropInRequest.threeDSecureRequest(threeDSecureRequest);
+        }
+
+        if (!mRequestedThreeDSecure && shouldRequestThreeDSecureVerification()) {
```
sestevens: Maybe instead of `mRequestedThreeDSecure` this could be called `mPerformedThreeDSecureVerification`? On a first read-through, I assumed this variable indicated whether the merchant had requested 3DS, but that seems to be covered by `shouldRequestThreeDSecureVerification()`.

sestevens: I also wonder whether `mRequestedThreeDSecure` could be checked _inside_ of `shouldRequestThreeDSecureVerification()`. I think that might be a little cleaner, but I'm not sure how much refactoring that would take since `shouldRequestThreeDSecureVerification()` lives in a base class.

Drop-In/src/main/java/com/braintreepayments/api/dropin/AddCardActivity.java
```diff
@@ -324,12 +334,15 @@ public void onSmsCodeSent(String enrollmentId, boolean smsRequired) {
     @Override
     public void onCancel(int requestCode) {
         if (requestCode == BraintreeRequestCodes.THREE_D_SECURE) {
+            mRequestedThreeDSecure = false;
```
scannillo: ^ I second Susan's input on the name of this variable being changed. This at first reads as that we are setting the merchant's request of 3DS to false.

Drop-In/src/main/java/com/braintreepayments/api/dropin/AddCardActivity.java
```diff
@@ -283,19 +286,26 @@ protected void createCard() {
...
         }
     }
 
     @Override
     public void onPaymentMethodNonceCreated(PaymentMethodNonce paymentMethod) {
-        mBraintreeFragment.sendAnalyticsEvent("sdk.exit.success");
-        finish(paymentMethod, null);
+        if (shouldRequestThreeDSecureVerification() && mDropInRequest.getThreeDSecureRequest() == null) {
+            ThreeDSecureRequest threeDSecureRequest = new ThreeDSecureRequest().amount(mDropInRequest.getAmount());
+            mDropInRequest.threeDSecureRequest(threeDSecureRequest);
```
sestevens: The documentation for `threeDSecureRequest` on `DropInRequest` says that if the merchant doesn't provide an amount, we'll default to the amount on `DropInRequest`. But here, we're only defaulting the amount if the merchant doesn't provide a `threeDSecureRequest` (i.e., the request object is null). If the `threeDSecureRequest` is present, but the _amount_ is null, we don't provide a default.

@scannillo and I want to propose this alternative:
```
    @Override
    public void onPaymentMethodNonceCreated(PaymentMethodNonce paymentMethod) {
        if (!mRequestedThreeDSecure && shouldRequestThreeDSecureVerification()) {
            if (mDropInRequest.getThreeDSecureRequest() == null) {
                mDropInRequest.threeDSecureRequest(new ThreeDSecureRequest());
            }

            if (mDropInRequest.getThreeDSecureRequest().getAmount() == null) {
                mDropInRequest.getThreeDSecureRequest().amount(mDropInRequest.getAmount());
            }

            mDropInRequest.getThreeDSecureRequest().nonce(paymentMethod.getNonce());
            ThreeDSecure.performVerification(mBraintreeFragment, mDropInRequest.getThreeDSecureRequest());
            mRequestedThreeDSecure = true;
        } else {
            mBraintreeFragment.sendAnalyticsEvent("sdk.exit.success");
            finish(paymentMethod, null);
        }
  }
```

Drop-In/src/test/java/com/braintreepayments/api/dropin/DropInRequestUnitTest.java
```diff
@@ -41,6 +44,29 @@ public void includesAllOptions() {
         PayPalRequest paypalRequest = new PayPalRequest("10")
                 .currencyCode("USD");
 
+
```
scannillo: extra new line

Demo/src/androidTest/java/com/braintreepayments/demo/test/utilities/CardNumber.java
```diff
@@ -20,6 +20,9 @@
     public static final String THREE_D_SECURE_MPI_LOOKUP_ERROR = "4000000000000085";
     public static final String THREE_D_SECURE_MPI_SERVICE_ERROR = "4000000000000093";
 
+    public static final String THREE_D_SECURE_2_FRICTIONLESS_VERIFICATON = "4000000000001000";
+    public static final String THREE_D_SECURE_2_CHALLENGE_VERIFICATON = "4000000000001091";
```
hollabaq86: both of these constants are Visa cards... is it worth changing one out for a test MC?

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -200,13 +200,22 @@ public void onAnimationFinished() {
 
     @Override
     public void onPaymentMethodNonceCreated(final PaymentMethodNonce paymentMethodNonce) {
-        if (!mRequestedThreeDSecure && paymentMethodNonce instanceof CardNonce &&
+        if (!mPerformedThreeDSecureVerification && paymentMethodNonce instanceof CardNonce &&
```
hollabaq86: I don't understand this logic, here, and I think it's the naming convention that's throwing me.  If we `performed` 3DS verification, why do they then call `shouldRequest` 3DS verification?

Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityPowerMockTest.java
```diff
@@ -2,12 +2,16 @@
 
 import com.braintreepayments.api.BraintreeFragment;
 import com.braintreepayments.api.Card;
+import com.braintreepayments.api.PayPal;
 import com.braintreepayments.api.ThreeDSecure;
 import com.braintreepayments.api.dropin.view.EditCardView;
 import com.braintreepayments.api.models.CardBuilder;
+import com.braintreepayments.api.models.CardNonce;
 import com.braintreepayments.api.models.Configuration;
+import com.braintreepayments.api.models.ThreeDSecureRequest;
```
hollabaq86: Q about where this should go. Do we arrange this section alphabetically, or should we move `ThreeDSecureRequest` so that it's in close to `ThreeDSecure` for contextual organization?

Drop-In/src/test/java/com/braintreepayments/api/dropin/DropInRequestUnitTest.java
```diff
@@ -41,6 +41,28 @@ public void includesAllOptions() {
...
                 .currencyCode("USD");
 
+        ThreeDSecureRequest threeDSecureRequest = new ThreeDSecureRequest()
+                .nonce("abc-123")
+                .versionRequested(ThreeDSecureRequest.VERSION_2)
+                .amount("10.00")
+                .email("tester@example.com")
+                .mobilePhoneNumber("3125551234")
+                .billingAddress(new ThreeDSecurePostalAddress()
+                        .givenName("Given")
```
hollabaq86: I like using `Given` and `Surname` as examples here ❤️

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
Epreuve: `mPerformedThreeDSecureVerification` is a boolean that is set to tru when a card is 3DS verified in order to keep from reverifying the same card after it's just been verified.  If for some reason the 3DS Flow does not complete (`onError` or `onCancel` are called), `handleThreeDSecureFailure()` is called and will set it back to false so the next attempt will again go through 3DS.

`shouldRequestThreeDSecureVerification()` on the other hand checks that `dropInRequest.requestThreeDSecureVerification()` was set to true, that `dropInRequest.amount()` is set to a value, and finally that the merchant is configured for 3DS.

Demo/src/androidTest/java/com/braintreepayments/demo/test/utilities/CardNumber.java
Epreuve: These should be fine as we're testing the 3DS response from cardinal which should not be card brand dependent.

PR #144, opened 2019-08-15
Add 3DSecure 2.0 snippet to readme

Epreuve:
Adds a 3DSecure 2.0 snippet to the README for reference on usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
 
 ### 3D Secure + Drop-in
 
-The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it with `DropInRequest.requestThreeDSecureVerification(true)` and set an amount. Then, create a ThreeDSecureRequest() object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,6 +64,43 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
 }
 ```
 
+### 3D Secure + Drop-in
+
+The new Drop-In supports 3D-Secure verification. If you have enabled 3D-Secure in the control panel, enable it with `DropInRequest.requestThreeDSecureVerification(true)` and set an amount. Then, create a ThreeDSecureRequest() object, setting as many fields on it as possible; the more fields that are set, the less likely it is that a user will be be presented with a challenge. Make sure to attach this object to the BTDropInRequest before use.
```
hollabaq86:
```suggestion
The new Drop-In supports 3D-Secure verification. Assuming you have [3D-Secure configured](https://developers.braintreepayments.com/guides/3d-secure/configuration) for your account, enable it in your client with `DropInRequest.requestThreeDSecureVerification(true)` and set an amount. Then, create a ThreeDSecureRequest() object, setting `ThreeDSecurePostalAddress` and `ThreeDSecureAdditionalInformation` fields where possible; the more fields that are set, the less likely a user will be presented with a challenge. For more information, check our [3D Secure Migration Guide](https://developers.braintreepayments.com/guides/3d-secure/migration/android/v3#getting-ready-for-3ds-2). Make sure to attach this object to the `BTDropInRequest` before use.
```

hollabaq86: Some clarifying edits.

README.md
```diff
@@ -64,6 +64,43 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+    .streetAddress("555 Smith St")
+    .extendedAddress("#2")
+    .locality("Chicago")
+    .region("IL")
+    .postalCode("12345")
+    .countryCodeAlpha2("US");
+
+ThreeDSecureRequest threeDSecureRequest = new ThreeDSecureRequest()
+    .amount("1.00")
+    .versionRequested(Settings.getThreeDSecureVersion(this))
```
sestevens: `Settings.getThreeDSecureVersion(this)` is specific to our demo app, right? Not something a merchant would necessarily have? I wonder if it would make more sense to hard-code this like the other values.

sestevens:
```suggestion
    .versionRequested("2")
```

billwerges: We have a constant defined for this that should be used, `ThreeDSecureRequest.VERSION_2`

PR #146, opened 2019-08-19
Deprecate DropInRequest `amount` field

sestevens:
## Summary
Deprecating `DropInRequest` `amount` field. Instead, `amount` should be set on `ThreeDSecureRequest`.

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
@@ -66,10 +66,13 @@ public class DropInRequest implements Parcelable {
     }
 
     /**
+     * Deprecated. Use {@link ThreeDSecureRequest#amount(String)}
+     *
      * This method is optional. Amount is only used for 3D Secure verifications.
      *
      * @param amount Amount of the transaction.
      */
+    @Deprecated
     public DropInRequest amount(String amount) {
         mAmount = amount;
         return this;


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+
+* Deprecate `amount` property on `DropInRequest`
```
billwerges: The Android Drop-In release script updates the CHANGELOG file programmatically, so adding this now will cause a duplicate entry.

sestevens: Oh right, I forgot about that. I think we should either update the iOS release script to do the same or remove that part of the Android script so that we're consistent across repos.

scannillo: Voting to remove that portion of the Android release script after discussions in stand up.

sestevens:
> If we're marking this property as deprecated, we should probably update this method so it doesn't rely on it:
> https://github.com/braintree/braintree-android-drop-in/blob/136a78462da85d0cedeb9588508da9fa7fa3616d/Drop-In/src/main/java/com/braintreepayments/api/dropin/BaseActivity.java#L67

So, I don't think we can remove the usage of the `amount` field on `DropInRequest` entirely, because we're still allowing merchants to set the amount there instead of on `ThreeDSecureRequest`. But this brings up a good point -- I think we should be checking that the amount is either set on `DropInRequest` OR `ThreeDSecureRequest` in this method.

I updated the `shouldRequestThreeDSecureVerification` method [here](https://github.com/braintree/braintree-android-drop-in/pull/146/files#diff-72b54a560c854f8748543de9e98ae741) to reflect that. Lemme know if that makes sense.

@Epreuve Would love to get your 👁 on this too.

Issue #152, opened 2019-08-29
Use a double for amount for 3D Secure 2.0


dadino:
### General information

* SDK/Library version: 4.3.0
* Environment: Sandbox or Production
* Android Version and Device: 5+

### Issue description

We wasted 2 days trying to understand why the credit card flow was not working. When clicking "Add card" ("Aggiungi carta" in Italian), the button would transform to a progressbar and then go back to button, no message displayed. In our logs we had the Paypal-debug-id, that from my understanding is something to pass to the assistance, no other useful informations.

Turns out the problem was the amount field, used for 3D Secure 2.0. The library asks for a String, which is strange, since it only accept a double with a dot (.) separator. We're an Italian company and in our Locale the separator is a comma (,) so our initial implementation was not working. Nowhere it is specified that it needs to be a dot.

All this could be easily skipped by requiring a double instead of a String. Or at least have a meaningful error message.

This is our current implementation, for anyone who'll need it:

    DecimalFormatSymbols formatSymbols = DecimalFormatSymbols.getInstance();
    formatSymbols.setDecimalSeparator('.');
    final String amount = new DecimalFormat("#.00", formatSymbols).format(product().getPrice());

scannillo:
Hello @dadino . Thanks for pointing this out! We will update the docstring on `ThreeDSecureRequest.amount` to be more clear.

Just to confirm, are you generating your client token or tokenization key using the proper `merchantID` configured with your appropriate country/ currency?

dadino:
> Just to confirm, are you generating your client token or tokenization key using the proper `merchantID` configured with your appropriate country/ currency?

We just checked and our sandbox is configured with a US timezone and EUR currency. Was the error due to this? What will happen when I use the dot String in the live server that has the Italian timezone?

scannillo:
Ok, glad we checked! Definitely make sure your sandbox is configured with the appropriate timezone and currency.

You should be fine to pass in your regular currency format.

PR #153, opened 2019-08-29
Update DropInRequest Amount Docstring

scannillo:
#### Summary of changes

 - Updated `DropInRequest.amount` docstring to include instructions for merchants to properly format the amount string.
- Solves this [issue](https://github.com/braintree/braintree-android-drop-in/issues/152).

 #### Changelog

Not necessary for this

 #### Questions

- There are processor limits on amount size, but I think it suffices that it live on the docs for `Transaction.sale()`'s `amount` param [here](https://developers.braintreepayments.com/reference/request/transaction/sale/ruby#amount).

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
@@ -71,7 +71,7 @@ public class DropInRequest implements Parcelable {
      * This method is optional. Amount is only used for 3D Secure verifications.
      *
      * This value must be a non-negative number and must match the currency format of the merchant account.
-     * It can only contain numbers and can optionally include one decimal point with exactly 2 decimal place precision (ex: x.xx).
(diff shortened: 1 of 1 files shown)


Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
```diff
@@ -70,8 +70,8 @@ public DropInRequest tokenizationKey(String tokenizationKey) {
      *
      * This method is optional. Amount is only used for 3D Secure verifications.
      *
-     * Amount must be a non-negative number, may optionally contain exactly 2 decimal places
-     * separated by '.', optional thousands separator ',', limited to 7 digits before the decimal point.
+     * This value must be a non-negative number and must match the currency format of the merchant account.
+     * It can only contain numbers and can optionally include one decimal point with exactly 2 decimal place precision (ex: x.xx).
```
sestevens:
```suggestion
     * It can only contain numbers and optionally one decimal point with exactly 2 decimal place precision (e.g., x.xx).
```
Small suggestion 🙂

PR #158, opened 2019-09-13
Fix on error add card activity

Epreuve:
### Summary of changes

 - Ensure that `finish(error)` is always called to avoid causing a hang when adding a card in some cases.

 ### Checklist

 - [x] Added a changelog entry


Drop-In/src/main/java/com/braintreepayments/api/dropin/AddCardActivity.java
```diff
@@ -374,9 +374,8 @@ public void onError(Exception error) {
             } else if (error instanceof DownForMaintenanceException) {
                 mBraintreeFragment.sendAnalyticsEvent("sdk.exit.server-unavailable");
             }
-
-            finish(error);
         }
+        finish(error);
```
quinnjn: We wouldn't want to finish this activity on any error. Field errors are processed on L361 for example which causes a state change in this activity.

Epreuve: I see now.  This change would effectively change the activity making the feedback on the field annoyingly un actionable.  Let me take another wack at this then.

Demo/src/androidTest/java/com/braintreepayments/demo/test/DropInTest.java
```diff
@@ -121,6 +121,9 @@ public void performsThreeDSecure2FrictionlessVerification() {
         onDevice(withTextStartingWith("created")).check(text(endsWith("authorized")));
     }
 
+    @Ignore("Test fails due to Cardinal Activity changes that resulted in the authorization code " +
```
quinnjn: Nitpick: I think these tests are unrelated to this PR. We should commit this on master first and then rebase this PR.

Epreuve:
> Why are there no tests for this?

Fixed!  Good call out.

Drop-In/src/test/java/com/braintreepayments/api/dropin/view/EditCardViewUnitTest.java
```diff
@@ -353,6 +354,26 @@ public void setVisibility_toVisibleFocusesCvvWhenOptional() {
...
+    public void isEditCardError_returnsTrue_whenCreditCardError() {
+        ErrorWithResponse error = new ErrorWithResponse(422, stringFromFixture("responses/credit_card_error_response.json"));
+
+        assertTrue(mView.isEditCardError(error));
+    }
+
+    @Test
+    public void isEditCardError_returnsFalse_whenErrorIsNotCreditCardOrUnionPay() {
+        ErrorWithResponse error = new ErrorWithResponse(403, new JSONObject().toString());
+        assertFalse(mView.isEditCardError(error));
```
sestevens:
```suggestion

        assertFalse(mView.isEditCardError(error));
```

Drop-In/src/main/java/com/braintreepayments/api/dropin/view/EditCardView.java
```diff
@@ -114,6 +114,11 @@ public void setMaskCvv(boolean mask) {
         mCardForm.maskCvv(mask);
     }
 
+    public boolean isEditCardError(ErrorWithResponse errors) {
+        return errors.errorFor("unionPayEnrollment") != null ||
```
hollabaq86: why do we need to check for UnionPay specifically here?

Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityUnitTest.java
```diff
@@ -546,6 +547,21 @@ public void onPaymentMethodNonceCreated_sendsAnalyticsEvent() throws JSONExcepti
         verify(mActivity.braintreeFragment).sendAnalyticsEvent("sdk.exit.success");
     }
 
+    @Test
+    public void nonFormFieldError_callsFinishWithError() {
+        setup(mock(BraintreeFragment.class));
+
+        ErrorWithResponse error = new ErrorWithResponse(422, stringFromFixture("responses/three_d_secure/authentication_response_with_error.json"));
```
scannillo: Why is this JSON specific to 3DS authentication?

crookedneighbor: If it is specific to 3ds, perhaps the test name should reflect it.

Drop-In/src/main/java/com/braintreepayments/api/dropin/view/EditCardView.java
Epreuve: In the case of union pay, there is no `creditCard` key and all errors are nested under the `unionPayEnrollment` key.  This checks both to determine if there are errors in either case.

Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityUnitTest.java
Epreuve: The code itself isn't specific to 3D Secure, so the test name reflects that.  However, the case that prompted the bug fix did have to do with 3D Secure, which is the reason for the name of the fixture.

However, the way the code is written, an error could come from a non 3D Secure case and flow through this code path as well, so we did not want to give the impression that _only_ 3DS errors would make it here.

scannillo: Curious if the fixture should uphold for a more generic error case as well? Or if we should add another.

Epreuve: We opted to rewrite the test to be very specific in testing the error payload, and avoid any unnecessary 3D Secure details in the JSON blob.

scannillo: This looks good to me! 👍

PR #182, opened 2020-06-08
Add 3DS support for non-network tokenized Google Pay cards

billwerges:
### Summary of changes

 - Runs 3DS verification for non-network tokenized Google Pay cards if 3DS is enabled as well.

 ### Checklist

 - [X] Added a changelog entry

### Authors

- @billwerges 
- @demerino

PR #183, opened 2020-06-09
Fix test expiry date and refactor test client token fixtures

Epreuve:
### Summary of changes

 The expiry date helper const is no longer valid, and was preventing
tests from correctly running or testing the appropriate failures via
Robolectric.  Bumping this to a considerable date in the future should
be sufficient.

The refactor for the client token fixtures is due to a bugfix we
deployed [here](https://github.braintreeps.com/team-sdk/braintree-android/pull/446/files) that corrects an unintended behavior
where a merchant could pass a regular string as an auth.  This
reinforces the need to only pass encoded client token or UATs, or
tokenization keys, and rejects random strings as incorrectly formatted.

Since this branch is required to get https://github.com/braintree/braintree-android-drop-in/pull/182 passing, but wasn't specific to the work in that PR, it was moved to it's own branch and PR.  Ideally the two would be merged, but as both are fairly sizable, this makes them more easily reviewable.  Whether this PR is closed and the work merged into 182, or, this is merged directly into master first (though it relies on the braintree_android bump in 182 to pass tests) is up for consideration, either would technically work.

 ### Checklist

- @Epreuve 
- @sshropshire

### DIFF
--- a/Drop-In/src/sharedTest/java/com.braintreepayments.api.test/ExpirationDate.java
+++ b/Drop-In/src/sharedTest/java/com.braintreepayments.api.test/ExpirationDate.java
@@ -2,5 +2,5 @@ package com.braintreepayments.api.test;
 
 public class ExpirationDate {
 
-    public static final String VALID_EXPIRATION = "1219";
+    public static final String VALID_EXPIRATION = "1229";
 }
--- a/Drop-In/src/sharedTest/java/com.braintreepayments.api.test/ExpirationDate.java
+++ b/Drop-In/src/sharedTest/java/com.braintreepayments.api.test/ExpirationDate.java
@@ -1,6 +1,12 @@
 package com.braintreepayments.api.test;
 
+import java.util.Calendar;
+
 public class ExpirationDate {
 
-    public static final String VALID_EXPIRATION = "1229";
+    public static final String VALID_EXPIRATION = "12" + validExpirationYear();
+
+    public static String validExpirationYear() {
+        return String.valueOf(Calendar.getInstance().get(Calendar.YEAR) + 1).substring(2);
+    }
 }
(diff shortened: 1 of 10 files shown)


Drop-In/src/sharedTest/java/com.braintreepayments.api.test/ExpirationDate.java
```diff
@@ -2,5 +2,5 @@
 
 public class ExpirationDate {
 
-    public static final String VALID_EXPIRATION = "1219";
+    public static final String VALID_EXPIRATION = "1229";
```
billwerges, line 5: This is far in the future, but it would probably be best to programmatically generate a date based on the current year.

Drop-In/src/test/java/com/braintreepayments/api/test/UnitTestFixturesHelper.java
```diff
@@ -26,4 +28,8 @@ public static String stringFromFixture(String filename) {
             throw new RuntimeException(e);
         }
     }
+
+    public static String base64EncodedClientTokenFromFixture(String filename) {
+                return Base64.getEncoder().encodeToString(stringFromFixture(filename).getBytes());
```
scannillo, line 33:
```suggestion
    return Base64.getEncoder().encodeToString(stringFromFixture(filename).getBytes());
```

Drop-In/src/sharedTest/java/com.braintreepayments.api.test/ExpirationDate.java
scannillo: +1. We do this it dynamically in the base SDK 
https://github.braintreeps.com/team-sdk/braintree-android/blob/master/TestUtils/src/main/java/com/braintreepayments/testutils/ExpirationDateHelper.java

Epreuve: Yeah, I had considered that.  It would be a bit more work since that const is used quite a few places, but if we think it's worth doing here, I'm happy to do some reworking there, too.  I gambled originally on getting the tests working, and potentially adding tasks for improvements, given these have been failing since ~December. 😱

scannillo: That's also fine! Creating a ticket for it works 👍

PR #189, opened 2020-08-11
Update Card Form Version

sarahkoop:
### Summary of changes

 - Bump `card-form` to version `5.0.0`

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sshropshire @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+* Bump card-form version to 5.0.0
+
 ## 4.6.0
 * Bump braintree_android version to 3.11.1
 * Bump card-form version to 4.3.0 (Updates Card icons)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Braintree Android Drop-In Release Notes
 
 ## unreleased
-* Bump card-form version to 5.0.0
+* Breaking Changes
+  * Bump card-form version to 5.0.0 (removes support for CardIO, a no longer maintained framework)
 
 ## 4.6.0
 * Bump braintree_android version to 3.11.1
--- a/Demo/src/androidTest/java/com/braintreepayments/demo/test/CardholderNameDropInTest.java
+++ b/Demo/src/androidTest/java/com/braintreepayments/demo/test/CardholderNameDropInTest.java
@@ -28,7 +28,7 @@ public class CardholderNameDropInTest extends TestHelper {
     @Test(timeout = 60000)
     public void cardholderName_whenDisabled_isHidden() {
         setCardholderNameStatus(CardForm.FIELD_DISABLED);
-        onDevice(withText("Add Payment Method")).waitForExists().waitForEnabled().perform(click());
+        onDevice(withText("Add Payment Method")).waitForExists(10000).waitForEnabled().perform(click());
         onDevice(withText("Credit or Debit Card")).perform(click());
         onDevice(withText("Card Number")).perform(setText(VISA));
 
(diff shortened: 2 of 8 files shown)


Drop-In/src/main/java/com/braintreepayments/api/dropin/AddCardActivity.java
```diff
@@ -378,23 +378,9 @@ public void onError(Exception error) {
         }
     }
 
-    @Override
-    public boolean onCreateOptionsMenu(Menu menu) {
```
scannillo, line 382: Is deleting this public method technically a breaking change?

sshropshire: In this case it should be fine, `onCreateOptionsMenu` is a method on the `AppCompatActivity` super class.

scannillo: Yeh - I wasn't sure if since we had some custom logic inside our override that removing our override was technically a breaking change.

sshropshire: Yeah I know removing CardIO was a breaking change for `android-card-form`. I wonder if it makes sense to major version bump drop-in since CardIO is no longer there 🤔

scannillo: Hm - I think since it isn't making incompatible API changes for braintree-android-drop-in users, it doesn't qualify for a major bump.

sestevens: I dunno - it's removing pretty significant functionality. I think it would qualify for a major version bump.

Demo/src/androidTest/java/com/braintreepayments/demo/test/CardholderNameDropInTest.java
```diff
@@ -28,7 +28,7 @@ public void setup() {
     @Test(timeout = 60000)
     public void cardholderName_whenDisabled_isHidden() {
         setCardholderNameStatus(CardForm.FIELD_DISABLED);
-        onDevice(withText("Add Payment Method")).waitForExists().waitForEnabled().perform(click());
+        onDevice(withText("Add Payment Method")).waitForExists(20000).waitForEnabled(20000).perform(click());
```
crookedneighbor, line 31: just curious, is there a reason we need to specify a wait time now?

Drop-In/src/main/java/com/braintreepayments/api/dropin/AddCardActivity.java
crookedneighbor: It's definitely a breaking change to drop Card.io, so whether this particular change is breaking or not shouldn't matter. We will need to do a major version bump regardless.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+* Bump card-form version to 5.0.0
```
crookedneighbor, line 4: we should add a breaking changes section and include that support for Card.io was dropped

Demo/src/androidTest/java/com/braintreepayments/demo/test/CardholderNameDropInTest.java
sshropshire: Yeah the `waitForExists()` method calls `waitForExists(5000)` under the hood. Increasing the timeout out to 20 seconds made the tests more stable for some reason.

CHANGELOG.md
sshropshire: 👍

Drop-In/src/main/java/com/braintreepayments/api/dropin/AddCardActivity.java
sestevens: Agreed.

crookedneighbor: I'm not too familiar with the changes here, but does anything change with the merchant with this removal? If so, it should be listed in the changelog under the breaking changes section.

sshropshire: Yeah, there is a superclass implementation that provides a default home button in the Android [Action Bar](https://developer.android.com/training/appbar). It's [Activity#createOptionsMenu](https://developer.android.com/reference/android/app/Activity.html#onCreateOptionsMenu(android.view.Menu)).

We were adding a _Scan Card_ icon as an additional menu item to activate card.io. We could mention that this button was removed, but merchants shouldn't have to take additional action here.

sshropshire:
@demerino we discussed releasing as v5.0.0 since we're technically removing functionality. Everyone's leaning toward a major version bump for the most part.

PR #193, opened 2020-08-20
Vault Venmo Option

sshropshire:
### Summary of changes

 - Add `vaultVenmo` option to `DropInRequest`

 ### Checklist

 - [x] Added a changelog entry

### Authors
- @sarahkoop 
- @sshropshire

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
@@ -28,6 +28,7 @@ public class DropInRequest implements Parcelable {
 
     private GooglePaymentRequest mGooglePaymentRequest;
     private PayPalRequest mPayPalRequest;
+
     private boolean mGooglePaymentEnabled = true;
     private boolean mMaskCardNumber = false;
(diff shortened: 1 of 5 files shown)


Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInRequest.java
```diff
@@ -214,6 +217,16 @@ public DropInRequest vaultCard(boolean defaultValue) {
         return this;
     }
 
+    /**
+     * @param defaultValue the default value used to determine if Drop-in should vault the customer's venmo payment method. Must be set to `false` when using a client token without a `customerId`.
```
scannillo, line 221:
```suggestion
     * @param defaultValue the default value used to determine if Drop-in should vault the customer's Venmo payment method. Must be set to `false` when using a client token without a `customerId`.
```

scannillo, line 221:
```suggestion
     * @param defaultValue the default value used to determine if Drop-in should vault the customer's Venmo payment method. Must be set to `false` when using a client token without a `customerId`.
```

PR #200, opened 2020-10-20
Upgrade Braintree Android to 3.14.2

sarahkoop:
### Summary of changes

 - Bump braintree_android version to 3.14.2

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sshropshire 
- @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## unreleased
 
-* Bump braintree_android version to 3.14.2
+* Bump braintree_android version to 3.14.2 (fixes [#305](https://github.com/braintree/braintree_android/issues/305))
+* Upgrade gradle build tools to 4.1.0
 
 ## 5.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 
-* Bump braintree_android version to 3.14.2 (fixes [#305](https://github.com/braintree/braintree_android/issues/305))
+* Bump braintree_android version to 3.14.2 (fixes [#197](https://github.com/braintree/braintree-android-drop-in/issues/197))
 * Upgrade gradle build tools to 4.1.0
 
 ## 5.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+
+* Bump braintree_android version to 3.14.2
+
 ## 5.0.0
 
 * Add `vaultVenmo` option to `DropInRequest`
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+
+* Bump braintree_android version to 3.14.2
```
hollabaq86, line 5: should we add a note that this update fixes the issue with the cardinal SDK?

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.1'
+        classpath 'com.android.tools.build:gradle:4.1.0'
```
scannillo, line 11: Not sure if this needs to also go in the CHANGELOG?

scannillo (approved):

build.gradle
hollabaq86, line 11: should we call out this bump in dependencies in the CHANGELOG?

sshropshire: We could this may end up changing a good bit though. I'd put this in the same category as our more traditional gradle dependencies. I wonder if we should start logging dependency updates though overall?

CHANGELOG.md
sshropshire: Most def yeah. We closed the issue as a duplicate and re-directed everyone to `braintree-android` for status updates, but we can def reference it here.

build.gradle
sshropshire: Yeah Sammy had a similar question, maybe it makes sense to note dependency updates in the CHANGELOG going forward. It could bring dependencies more to the forefront of our process.

CHANGELOG.md
scannillo: I think in general we try to keep the CHANGELOG entries specific to changes in this SDK. Merchants can reference the CHANGELOG of the other repo to see what changed there.

scannillo: I see the concern though since the Cardinal issue was pretty severe

sarahkoop: We've added a link back to the original drop in issue for consistency. That issue then links to the currently open issue in braintree-android, so anyone interested can find the full conversation.

sarahkoop:
retest this please

crookedneighbor (approved):

build.gradle
crookedneighbor: Does upgrading the gradle version have any impact on the merchant integrating with Drop-in?

sshropshire: It shouldn't. The Android Gradle Plugin is basically gradle's link into Android studio. Most merchants are compiling with the AAR library hosted by Nexus Sonatype.

PR #202, opened 2020-10-22
Disable Strict Mode

sarahkoop:
### Summary of changes

 - Temporarily disable strict mode in demo application

 ### Checklist

### Authors

- @sshropshire 
- @sarahkoop

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/DemoApplication.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/DemoApplication.java
@@ -12,10 +12,11 @@ import retrofit.RestAdapter;
 public class DemoApplication extends Application {
 
     private static ApiClient sApiClient;
+    private static final boolean STRICT_MODE_ENABLED = false;
 
     @Override
     public void onCreate() {
-        if (BuildConfig.DEBUG) {
+        if (STRICT_MODE_ENABLED) {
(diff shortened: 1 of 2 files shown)


demerino (approved):

sestevens (approved):
Looks good! I assume we're leaving this code here so that it's easier to re-enable later? Should we add a comment or something explaining that?

I'd also be OK with just removing the strict mode code.

sshropshire:
@sestevens yup! Looking at the [dev docs](https://developer.android.com/reference/android/os/StrictMode) Strict Mode may still be useful post v4, but right now it's a bit of a hindrance. I can add a comment to `re-add` after we refactor to use a PayPal analytics module or switch to Room.

sshropshire:
@sestevens on second thought i'm down to remove haha. Just read this on the site:

> If you find violations that you feel are problematic, there are a variety of tools to help solve them: threads, Handler, AsyncTask, IntentService, etc. But **don't feel compelled to fix everything that StrictMode finds**. In particular, many cases of disk access are often necessary during the normal activity lifecycle. Use StrictMode to find things you did by accident. Network requests on the UI thread are almost always a problem, though.

So we could leave it up to the future Us's to determine if it's worth adding back.

scannillo (approved):

PR #204, opened 2020-11-06
Convert Test Fixtures to Single Kotlin Fixtures Object

sarahkoop:
### Summary of changes

 - Move JSON test fixtures to single Kotlin file to align with fixtures changes coming in braintree-android

 ### Checklist

### Authors

- @sarahkoop


Drop-In/src/test/java/com/braintreepayments/api/dropin/BaseActivityUnitTest.java
```diff
@@ -131,7 +131,7 @@ public void getBraintreeFragment_throwsAnExceptionForEmptyAuthorization() throws
     @Test
     public void getBraintreeFragment_setsClientTokenPresentWhenAClientTokenIsPresent() throws InvalidArgumentException {
         setup(new DropInRequest()
-                .clientToken(base64EncodedClientTokenFromFixture("client_token.json"))
+                .clientToken(base64EncodedClientTokenFromFixture(Fixtures.CLIENT_TOKEN))
```
sarahkoop, line 134: The result of this Base64 encoding method on the client token fixture could also be added directly as a `BASE64_CLIENT_TOKEN` fixture to the `Fixtures` file. That would allow the `base64EncodedClientTokenFromFixture` method and the `UnitTestFixturesHelper` to be removed all together. Do we think it is valuable to keep this helper method, or should the client token be encoded and added as a fixture directly?

scannillo: Hm - I can see either way. Though I lean toward leaving as-is allows the JSON to be human readable.

But since we're only using `base64EncodedClientTokenFromFixture` on the `client_token.json` and no other variants, we could do away with it and include the JSON value as a comment next to the fixture.

scannillo (approved):
Looks good to me!

scannillo:
Question - so if we need to add new test files to the Android SDKs at this point, should we opt for Kotlin as often as possible? Sort of like how any new test files in the iOS SDK - we're using Swift?

Drop-In/src/test/java/com/braintreepayments/api/dropin/BaseActivityUnitTest.java
sestevens: I like having the CLIENT_TOKEN fixture consist of JSON (how it is now) instead of hard-coding the base64 encoded string. Like Sammy said, it's easier to read. It also makes it easier to add additional fields to the JSON in the future.

sestevens (approved):
Looks good to me! 🌷

sestevens:
> Question - so if we need to add new test files to the Android SDKs at this point, should we opt for Kotlin as often as possible? Sort of like how any new test files in the iOS SDK - we're using Swift?

I'd vote yes for this approach! It lets us see how our APIs translate into Kotlin, so that we can make sure the Java code we're writing is Kotlin-friendly.

sarahkoop:
> > Question - so if we need to add new test files to the Android SDKs at this point, should we opt for Kotlin as often as possible? Sort of like how any new test files in the iOS SDK - we're using Swift?
> 
> I'd vote yes for this approach! It lets us see how our APIs translate into Kotlin, so that we can make sure the Java code we're writing is Kotlin-friendly.

We haven't been converting the tests themselves or writing new test files in Kotlin yet, but I think that is a good idea moving forward.

Issue #205, opened 2020-12-04
Vaulted payment methods are not filtered out from disablePayPal(), disableCard(), etc


davideavagliano:
### General information

SDK/Library version: all
Environment: all

### Issue description

When any one of the following options is used for a `DropInRequest`:
 - disableGooglePayment()
 - disablePayPal()
 - disableVenmo()
 - disableCard()

The `DropInActivity` filters out disabled payment modes only from `SupportedPaymentMethodsAdapter`, without doing the same also for `VaultedPaymentMethodsAdapter`.

This means that if I disable PayPal from within a particular `DropInRequest`, I won't be able to add a NEW Paypal payment method, but I will still be able to select a PayPal payment method I've used before.

davideavagliano:
🔝

PR #206, opened 2020-12-04
Filter Vaulted Payment Methods

sshropshire:
### Summary of changes

 - Filter Vaulted Payment Methods based on `DropInRequest` (fixes [#205](https://github.com/braintree/braintree-android-drop-in/issues/205))

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop 
- @sshropshire


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+
+* Filter Vaulted Payment Methods based on `DropInRequest` (fixes [#205](https://github.com/braintree/braintree-android-drop-in/issues/205))
```
sestevens, line 5:
```suggestion
* Filter Vaulted Payment Methods based on `DropInRequest` (fixes #205)
```

As long as the issue is in the same repo, GitHub should make the link for you automatically!

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
```diff
@@ -0,0 +1,67 @@
...
+                if (supportedCardTypes.size() > 0) {
+                    shouldAddPaymentMethod = true;
+                }
+            }
+
+            if (paymentMethodNonce instanceof GooglePaymentCardNonce && googlePayEnabled) {
+                if (dropInRequest.isGooglePaymentEnabled()) {
+                    shouldAddPaymentMethod = true;
+                }
+            }
```
billwerges, line 52: To make this loop a little more efficient, do we need to check the `paymentMethodNonce` type against every payment method class even if we've already found a match? 

If possible we could use a `switch` statement to check against the specific type of payment method, or just convert this into an `if-else` block so that we can exit early.

CHANGELOG.md
sshropshire: Good call. I always go back and forth with that haha. I like the less verbose way, it's a lot cleaner.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sshropshire: Yeah agreed, we can do an if-else here to make it more efficient.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/VaultedPaymentMethodsAdapter.java
```diff
@@ -20,12 +23,15 @@
 public class VaultedPaymentMethodsAdapter extends RecyclerView.Adapter<VaultedPaymentMethodsAdapter.ViewHolder> {
 
     private PaymentMethodNonceCreatedListener mSelectedListener;
-    private List<PaymentMethodNonce> mPaymentMethodNonces;
+    private AvailablePaymentMethodNonceList mPaymentMethodNonces;
 
-    public VaultedPaymentMethodsAdapter(PaymentMethodNonceCreatedListener listener,
-                                        List<PaymentMethodNonce> paymentMethodNonces) {
+    public VaultedPaymentMethodsAdapter(PaymentMethodNonceCreatedListener listener) {
```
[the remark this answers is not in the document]
sshropshire: Agreed yeah we can do this with a minor version bump with some refactoring. It's less clean but it'll work until the next major version bump of drop-in.

demerino (approved):
How would the payment methods get added in the first place if they are disabled?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+
+* Filter Vaulted Payment Methods based on `DropInRequest` (fixes #205)
```
sestevens, line 5: Is it worth being a little more specific here? Something like:
```suggestion
* Fix issue where payment methods disabled on `DropInRequest` were displayed if previously vaulted (fixes #205)
```

sshropshire:
@demerino I actually thought about that for a second too. I guess if a payment method had been vaulted previously and the same payment method is disabled by the merchant, it could happen. It may be a bit of an edge case.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
```diff
@@ -0,0 +1,59 @@
...
+                shouldAddPaymentMethod = true;
+            } else if (paymentMethodNonce instanceof VenmoAccountNonce && dropInRequest.isVenmoEnabled() && configuration.getPayWithVenmo().isEnabled(context)) {
+                shouldAddPaymentMethod = true;
+            } else if (paymentMethodNonce instanceof CardNonce && dropInRequest.isCardEnabled()) {
+                Set<String> supportedCardTypes =
+                        new HashSet<>(configuration.getCardConfiguration().getSupportedCardTypes());
+                if (!unionpaySupported) {
+                    supportedCardTypes.remove(PaymentMethodType.UNIONPAY.getCanonicalName());
+                }
+                if (supportedCardTypes.size() > 0) {
```
sestevens, line 39: We could probably use `empty` instead of checking the size here.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
```diff
@@ -0,0 +1,59 @@
...
+                    shouldAddPaymentMethod = true;
+                }
```
sestevens, line 41: Where did this logic around union pay come from? It seems a little odd, because as long as the merchant supports union pay **and** another card type, a vaulted union pay card could be displayed even if `unionpaySupported` is false.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
```diff
@@ -0,0 +1,59 @@
...
+
+    final private List<PaymentMethodNonce> items;
+
+    AvailablePaymentMethodNonceList(Context context, Configuration configuration, List<PaymentMethodNonce> paymentMethodNonces, DropInRequest dropInRequest, boolean googlePayEnabled, boolean unionpaySupported) {
+        items = new ArrayList<>();
+
+        for (PaymentMethodNonce paymentMethodNonce: paymentMethodNonces) {
+            boolean shouldAddPaymentMethod = false;
+
+            if (paymentMethodNonce instanceof PayPalAccountNonce && dropInRequest.isPayPalEnabled() && configuration.isPayPalEnabled()) {
```
sestevens, line 29: These long if statements are a little hard for me to read. Would it make sense to do something like:
```
if (paymentMethodNonce instanceof PayPalAccountNonce) {
    shouldAddPaymentMethod =  dropInRequest.isPayPalEnabled() && configuration.isPayPalEnabled();
} else if (paymentMethodNonce instanceof VenmoAccountNonce) {
    shouldAddPaymentMethod = dropInRequest.isVenmoEnabled() && configuration.getPayWithVenmo().isEnabled(context)
}
```

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sshropshire: Yeah I agree using the convenience methods can make the code more readable.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sshropshire: Yeah I agree it's a direct port of the [SupportedPaymentMethodsAdapter](https://github.com/braintree/braintree-android-drop-in/blob/master/Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/SupportedPaymentMethodsAdapter.java) logic. We put it in a data structure to decouple from Android. There wasn't an easy way to extract and reuse the functionality since one filters enums and the other filters nonce instances.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sshropshire: Def yeah this works too. It should for sure cut down on the column count and make more readable.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sarahkoop: We've updated this to use `isEmpty()` and flattened the if statement.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sarahkoop: Yeah the shorter If statements are more readable, we've made this change.

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
```diff
@@ -449,4 +436,35 @@ public void onVaultEditButtonClick(View view) {
...
+                this, mConfiguration, mDropInRequest, googlePayEnabled, mClientTokenPresent);
+        mVaultedPaymentMethodsView.setAdapter(vaultedPaymentMethodsAdapter);
+
+        if (mDropInRequest.isVaultManagerEnabled()) {
+            mVaultManagerButton.setVisibility(View.VISIBLE);
+        }
+
+        for (PaymentMethodNonce nonce : paymentMethodNonces) {
+            if (nonce instanceof CardNonce) {
+                mBraintreeFragment.sendAnalyticsEvent("vaulted-card.appear");
```
sestevens, line 465: I wonder if we should move this analytics event into `VaultedPaymentMethodsAdapter` since vaulted cards might not actually appear if cards aren't enabled by the merchant in the `DropInRequest`.

sestevens: Alternatively, we could move the construction of `AvailablePaymentMethodNonceList` into this method, and then only pass the available payment method nonces into `VaultedPaymentMethodsAdapter`. Then we would have access to the list of displayed vaulted cards in this method, and we could use that to determine whether or not to send this event.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sestevens: I'm having trouble wrapping my head around the logic in `SupportedPaymentMethodsAdapter`, but it seems like we shouldn't display union pay cards if union pay isn't supported. Is there a way to check if the nonce is a union pay nonce, and then only display it if union pay is supported?

For other cards, it seems like we should always display them as long as cards are enabled on `DropInRequest`. I suppose there's an edge case where a customer vaults a card, and then the merchant turns off support for that card in the Control Panel, but I don't think we should be handling that on the front-end.

sshropshire: Yeah this is a good call I don't think there's a true equivalent for that if-else clause in this data structure. Supported payment methods adapter is basically trying to determine if it should display `PaymentMethodType.UNKNOWN` which is basically the `Add Card` button. This would guard against Union Pay nonces being displayed iff Union Pay was a merchant's only supported card brand. 🤔

sshropshire: And I think we would have to use `UnionPay.fetchCapabilities()` to check if a Card Nonce is a Union Pay Nonce 😞

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
sshropshire: Yeah that analytics event is interesting we should preserve existing behavior. I think iterating over the filtered list makes a lot of sense.

Drop-In/src/main/java/com/braintreepayments/api/dropin/adapters/AvailablePaymentMethodNonceList.java
sshropshire: Yeah this is tough I think we would need a single network call per nonce to check if UnionPay I'm unaware of any method to check if it's UnionPay without making the round trip.

sarahkoop: We ended up simplifying this to only check if cards are enabled in the request and any supported card type exists. We can revisit UnionPay in the future if needed.

Drop-In/src/main/java/com/braintreepayments/api/dropin/DropInActivity.java
sarahkoop: We added a check in the `AvailablePaymentMethodNonceList` to see if the list has cards and only fire this event if it does.

sestevens (approved):
Looks good to me!

PR #210, opened 2021-01-21
Update Google Pay Icon

sshropshire:
### Summary of changes

 - Update Google Pay Icon to match Google style guidelines

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop 
- @sshropshire


hollabaq86 (approved):
Overrall LGTM. Do we no longer need to support xxxhpdi specifically? I don't see this icon being replaced

sshropshire:
> Overrall LGTM. Do we no longer need to support xxxhpdi specifically? I don't see this icon being replaced

That's a good call. We did provide xxxhdpi but for some reason it's the only asset with a decrease in file size. We're looking into it the command line tool we're using is changing some of the pngs from 8-bit to 32-bit color.

scannillo (approved):

PR #211, opened 2021-01-21
Update 3DS Version 

sarahkoop:
### Summary of changes

 - Update `three-d-secure` to `3.15.0` to align with core dependency
 - Fix rake task that was running all unit tests twice

 ### Checklist

### Authors

- @sarahkoop 
- @sshropshire

### DIFF
--- a/Drop-In/build.gradle
+++ b/Drop-In/build.gradle
@@ -60,7 +60,7 @@ android {
 dependencies {
     api 'com.braintreepayments.api:braintree:3.15.0'
     api 'com.braintreepayments:card-form:5.0.0'
-    api 'com.braintreepayments.api:three-d-secure:3.14.2'
+    api 'com.braintreepayments.api:three-d-secure:3.15.0'
 
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'com.google.android.gms:play-services-wallet:16.0.0'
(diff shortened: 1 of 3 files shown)


sestevens (approved):

Drop-In/build.gradle
```diff
@@ -60,7 +60,7 @@ android {
 dependencies {
     api 'com.braintreepayments.api:braintree:3.15.0'
     api 'com.braintreepayments:card-form:5.0.0'
-    api 'com.braintreepayments.api:three-d-secure:3.14.2'
+    api 'com.braintreepayments.api:three-d-secure:3.15.0'
```
sestevens, line 63: Should we call this out in the CHANGELOG?

sarahkoop: Typically we have only been calling out bumping `braintree_android` version in the CHANGELOG. This update is just to align the 3DS dependency with the core dependency that was bumped in the last release. We could add it as a separate CHANGELOG item though?

sestevens: Hmm... Is there a reason why we didn't bump the `three-d-secure` version when we bumped the `braintree` version? I don't have strong feelings about adding a CHANGELOG entry, but it seems like it couldn't hurt? 🤷‍♀️

sshropshire: Yeah tbh rushing to get the Hostname Verifier fix out was the reason. The two dependencies are fundamentally the same since the `three-d-secure` went virtually untouched in the update.

sestevens: Gotcha, that totally makes sense. Yeah, I'd say it's up to you guys if you think a CHANGELOG entry would be helpful.

crookedneighbor (approved):
Agree that a changelog entry would probably be appropriate here.

PR #215, opened 2021-02-18
add BraintreeBrowserSwitchActivity in manifest

xanscale:
this thing dropin lib can do itself


sestevens:
Hi @xanscale, thank you for opening this PR! This will be changing a bit in the next major version of Drop-in. We're removing `BraintreeBrowserSwitchActivity`, and instead you'll need to specify an activity you own. You can see more about this in our [migration guide](https://github.com/braintree/braintree_android/blob/master/v4_MIGRATION_GUIDE.md#browser-switch) for the Braintree Android SDK.

We don't have an exact timeline for a new major version of Braintree Drop-in, but development for it is underway.

PR #233, opened 2021-06-09
Add PaymentMethodClient and Use Core 4.1.0

sarahkoop:
### Summary of changes

 - Add `PaymentMethodClient` and associated exceptions/callbacks to Drop-in
 - Use `braintree_android` 4.1.0

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop 
- @sshropshire


Drop-In/src/androidTest/java/com/braintreepayments/api/PaymentMethodClientTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private Context context;
+
+    @Before
+    public void setUp() {
+        context = ApplicationProvider.getApplicationContext();
+    }
+
+    @Test(timeout = 10000)
+    public void getPaymentMethodNonces_andDeletePaymentMethod_returnsCardNonce() throws InterruptedException {
```
scannillo, line 36: Should we split these up? Have 1 test that is just asserting on that getPaymentMethodNonces returns a nonce, and another test for the deletion?

sshropshire: I think it makes sense for them to be seperate, but we would probably have to name the tests in a way that makes sure to run the `getPaymentMethodNonces` test before the `deletePaymentMethod` test.

Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
```diff
@@ -342,4 +341,36 @@ private void safelyCloseLoadingView() {
             mLoading.dismiss();
         }
     }
+
+    // NOTE: these are duplicated from PaymentMethodNonceInspector, which is package private
+    // we need getNonceDescription and getNonceTypeLabel for our integration test assertions,
```
sestevens, line 346: Shouldn't these be public so that merchants can access them? On iOS, they are available through `BTDropInResult`: https://github.com/braintree/braintree-ios-drop-in/blob/master/Sources/BraintreeDropIn/Public/BraintreeDropIn/BTDropInResult.h.

Drop-In/src/androidTest/java/com/braintreepayments/api/PaymentMethodClientTest.java
sestevens: I don't think we want one test to depend on another test being run first. But we could have one test that just tests `getPaymentMethodNonces` and another one that tests `deletePaymentMethod` and calls `getPaymentMethodNonces` as part of its set-up.

Drop-In/src/test/java/com/braintreepayments/api/PaymentMethodClientUnitTest.java
```diff
@@ -0,0 +1,401 @@
...
+    ArgumentCaptor<List<PaymentMethodNonce>> paymentMethodNoncesCaptor;
+
+    @Before
+    public void beforeEach() {
+        MockitoAnnotations.initMocks(this);
+
+        cardNonce = mock(CardNonce.class);
+        context = ApplicationProvider.getApplicationContext();
+
+        when(cardNonce.getString()).thenReturn("im-a-card-nonce");
```
sestevens, line 46: 😂

Drop-In/src/androidTest/java/com/braintreepayments/api/PaymentMethodClientTest.java
sshropshire: This is an odd test. The client token is generated with a completely random customerId, so I think the expectation is this test is for a new customer. With that assumption, we would have `createNonce -> assertGetNoncesWorks -> deleteNonce` and `createNonce -> deleteNonce -> assertNoNoncesArePresent`.

We would benefit from having the assertions isolated in their own tests, but the setup / teardown wouldn't be much different. I'm 50/50.

sestevens (approved):

Drop-In/src/androidTest/java/com/braintreepayments/api/PaymentMethodClientTest.java
scannillo: Gotcha - also agree with Susan. Whatever you decide works for me on this one 👍

scannillo: Thank you for providing the additional info

scannillo (approved):
🎸

PR #240, opened 2021-06-25
Bump card-form to 5.X.X (Support Discover 19 Digits)

sarahkoop:
### Summary of changes

 - Change auto-navigation on card entry to only auto-transition when card is valid and max length for that card type. This allows support for Discover cards 16-19 digits and aligns with [change in iOS Drop-in](https://github.com/braintree/braintree-ios-drop-in/pull/333).
 - This change requires android-card-form release with with [this PR](https://github.com/braintree/android-card-form/pull/108)

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityUnitTest.java
+++ b/Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityUnitTest.java
@@ -47,6 +47,7 @@ import static androidx.appcompat.app.AppCompatActivity.RESULT_FIRST_USER;
 import static androidx.appcompat.app.AppCompatActivity.RESULT_OK;
 import static com.braintreepayments.api.test.Assertions.assertIsANonce;
 import static com.braintreepayments.api.test.CardNumber.AMEX;
+import static com.braintreepayments.api.test.CardNumber.UNIONPAY_19;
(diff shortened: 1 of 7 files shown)


Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityUnitTest.java
```diff
@@ -258,15 +259,15 @@ public void editingAUnionPayCardNumberIsPossible() {
                 .clientToken(base64EncodedClientTokenFromFixture(Fixtures.CLIENT_TOKEN)));
         setup(httpClient);
 
-        setText(mAddCardView, R.id.bt_card_form_card_number, UNIONPAY_CREDIT);
+        setText(mAddCardView, R.id.bt_card_form_card_number, UNIONPAY_19);
```
sarahkoop, line 262: Updated to use a 19 digit Union Pay test card so that the auto-transition occurs for this test.

build.gradle
```diff
@@ -33,6 +33,7 @@ allprojects {
     repositories {
         google()
         jcenter()
+        mavenLocal()
```
sarahkoop, line 36: To be removed after card form is released and version is updated.

PR #241, opened 2021-06-30
Convert AddCardActivity to Fragments

sarahkoop:
### Summary of changes

 - WIP

 ### Checklist

 - [ ] Added a changelog entry

-


Drop-In/build.gradle
```diff
@@ -73,7 +73,8 @@ dependencies {
     api 'com.braintreepayments.api:card:4.1.0'
     api 'com.braintreepayments.api:data-collector:4.1.0'
     api 'com.braintreepayments.api:union-pay:4.1.0'
-    api 'com.braintreepayments:card-form:5.1.0'
+
+    api 'com.braintreepayments:card-form:5.1.2-SNAPSHOT'
```
sarahkoop, line 77: Changes to card form are drafted [here](https://github.com/braintree/android-card-form/pull/109), so we are using a local build for testing (CI will fail for now)

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
```diff
@@ -0,0 +1,137 @@
...
+    fun whenStateIsRESUMED_buttonTextIsAddCard() {
+        val scenario = FragmentScenario.launchInContainer(AddCardFragment::class.java, null, R.style.bt_drop_in_activity_theme)
+        scenario.moveToState(Lifecycle.State.RESUMED)
+
+        scenario.onFragment { fragment ->
+            fragment.dropInViewModel.setSupportedCardTypes(listOf(CardType.VISA))
+        }
+
+        onView(isRoot()).perform(waitFor(500))
+        onView(withId(R.id.bt_button)).check(matches(withText(R.string.bt_next)))
```
scannillo, line 42: Should the tests be using the R id values or hard coded string values 🤔 . The strings are definitely easier to read. Also wouldn't using the raw strings catch issues if for some the R id values changed - our tests would break to catch it?

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
```diff
@@ -0,0 +1,137 @@
...
+        scenario.onFragment { fragment ->
+            fragment.dropInViewModel.setSupportedCardTypes(listOf(CardType.VISA))
+            fragment.dropInViewModel.setCardTokenizationError(ErrorWithResponse.fromJson(Fixtures.CREDIT_CARD_EXPIRATION_ERROR_RESPONSE))
+
+            assertNull(fragment.cardForm.cardEditText.textInputLayoutParent?.error)
+        }
+    }
+
+    @Test
+    fun whenStateIsRESUMED_sendsAnalyticsEventToNotifyAddCardSelected() {
```
scannillo, line 67:
```suggestion
    fun whenStateIsRESUMED_andAddCardSelected_sendsAnalyticsEvent() {
```
So that the "expected behavior" is the last part of the test name

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
```diff
@@ -0,0 +1,137 @@
...
+    @Test
+    fun whenStateIsRESUMED_cardNumberFieldIsFocused() {
+        val scenario = FragmentScenario.launchInContainer(AddCardFragment::class.java, null, R.style.bt_drop_in_activity_theme)
+        scenario.moveToState(Lifecycle.State.RESUMED)
+
+        onView(withId(R.id.bt_card_form_card_number)).check(matches(isFocused()))
+    }
+
+    @Test
+    fun whenStateIsRESUMED_doesNotShowErrorForNonCardNumberError() {
```
scannillo, line 54:
```suggestion
    fun whenStateIsRESUMED_andNonCardNumberError_doesNotShowError() {
```

Same comment as below - keeping the naming format of tests more consistent

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
```diff
@@ -0,0 +1,137 @@
...
+            fragment.dropInViewModel.setSupportedCardTypes(listOf(CardType.VISA))
+            fragment.dropInViewModel.setCardTokenizationError(ErrorWithResponse.fromJson(Fixtures.CREDIT_CARD_ERROR_RESPONSE))
+
+            assertEquals(fragment.context?.getString(R.string.bt_card_number_invalid),
+                    fragment.cardForm.cardEditText.textInputLayoutParent?.error)
+        }
+    }
+
+    @Test
+    fun whenStateIsRESUMED_whenCardFromInvalid_showsButton() {
```
scannillo, line 126:
```suggestion
    fun whenStateIsRESUMED_whenCardFormInvalid_showsButton() {
```

scannillo: This test is a little confusing - which button are we looking for?

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
```diff
@@ -0,0 +1,380 @@
...
+        onView(isRoot()).perform(waitFor(500))
+
+        onView(withId(R.id.bt_card_form_expiration)).perform(typeText(ExpirationDate.VALID_EXPIRATION))
+        scenario.onFragment { fragment ->
+            assertFalse(fragment.cardForm.isValid)
+        }
+
+        onView(withId(R.id.bt_card_form_cardholder_name)).perform(typeText("Brian Tree"))
+        scenario.onFragment { fragment ->
+            assertTrue(fragment.cardForm.isValid)
```
scannillo, line 105: Would this test be more powerful in checking that the Cardholder name is require by not typing in "Brian Tree" and then asserting the CardForm is Not valid?

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
```diff
@@ -0,0 +1,380 @@
...
+        }
+    }
+
+    @Test
+    fun whenStateIsRESUMED_withoutSaveCardCheckBoxPropertiesOnDropInRequest_saveCardCheckboxCheckedAndHidden() {
```
scannillo, line 110: Nitpick - but test names could use more consistent format. It isn't a huge deal, but I think makes it easier to read through the tests.

Example:

```
whenStateIsRESUMED_whenDropInRequestRequiresCardholderName_... 
whenStateIsRESUMED_whenDropInRequestUsesDefaultSaveCardCheckboxProperties_...
```

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
```diff
@@ -0,0 +1,380 @@
...
+        val events = mutableListOf<Parcelable>()
+        scenario.onFragment { fragment ->
+            val activity = fragment.requireActivity()
+            val fragmentManager = fragment.parentFragmentManager
+            fragmentManager.setFragmentResultListener("BRAINTREE_EVENT", activity) { requestKey, result ->
+                events += result
+            }
+        }
+        assertEquals(0, events.size)
+        onView(withId(R.id.bt_button)).check(matches(isDisplayed()))
```
scannillo, line 269: Same question - what button is this?

Drop-In/src/main/java/com/braintreepayments/api/AddCardActivity.java
```diff
@@ -1,419 +0,0 @@
...
-import com.braintreepayments.api.dropin.R;
-import com.braintreepayments.cardform.view.CardForm;
-
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-
-import static android.view.View.GONE;
-import static android.view.View.VISIBLE;
-
-public class AddCardActivity extends BaseActivity implements AddPaymentUpdateListener {
```
scannillo, line 25: BYE

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -36,12 +41,14 @@
...
 
     private DropInViewModel dropInViewModel;
+    ActionBar actionBar;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.bt_drop_in_activity);
 
+
```
scannillo, line 51:
```suggestion
```

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -83,6 +90,12 @@ private void handleBraintreeEventBundle(Bundle bundle) {
             VaultedPaymentMethodSelectedEvent vaultedPaymentMethodSelectedEvent =
                     (VaultedPaymentMethodSelectedEvent) braintreeResult;
             onVaultedPaymentMethodSelected(vaultedPaymentMethodSelectedEvent.getPaymentMethodNonce());
+        } else if (braintreeResult instanceof AddCardEvent) {
+            showCardDetailsFragment(((AddCardEvent) braintreeResult).getCardNumber());
+        } else if (braintreeResult instanceof CardDetailsEvent) {
+            onCardDetailsEvent((CardDetailsEvent) braintreeResult);
+        } else if (braintreeResult instanceof EditCardNumberEvent) {
+            startAddCardFlow((EditCardNumberEvent) braintreeResult);
```
scannillo, line 98: The naming here for these `Event` classes threw me off a bit. After stepping through the flow on a simulator, it makes more sense:`CardDetailsEvent` represents _after_ card details have been collected, `AddCardEvent` represents _after_ card number is collected. 

Is this Android convention? If not, is there a way we could make this a bit more clear?

scannillo:
Should there be an `Edit` button in the Vault `Recent` section of the payment selection screen?

Drop-In/src/main/java/com/braintreepayments/api/DropInPaymentMethodType.java
```diff
@@ -72,7 +72,7 @@ public static DropInPaymentMethodType forType(PaymentMethodNonce paymentMethodNo
      * @return a {@link CardType[]} containing the mapped {@link CardType}s from the supplied
      * supportedCardTypes.
      */
-    public static CardType[] getCardsTypes(Set<String> supportedCardTypes) {
+    public static CardType[] getCardsTypes(List<String> supportedCardTypes) {
```
scannillo, line 75: Curious why the switch here. We want the order that the List provides?

scannillo:
I haven't looked at Android Drop In in a while so I forget ... BUT was the `Next` button the Add Card view always enabled? It looks like it always enabled, even if the card # is empty or invalid. Just curious here

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
sshropshire: Yeah we're doing it like this because most of the Espresso test runner tutorials use this method. I agree, it's probably better overall to go with the actual text. We really wanted to get everything modernized and broken up into fragments, then we'll probably refine the testing strategy a bit more.

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
sshropshire: I think that would be the submit button here yeah, we should clarify.

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sshropshire: Yeah we do have an assertion to assert that card form `isValid = false` in the block above.

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sshropshire: Yeah I agree. We ported most of the names from the original tests. They could use a second look.

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sshropshire: It's the submit button on this screen. It has a generic name because it's nested inside of a custom `AnimationButtonView`.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: We matched Event and Fragment prefixes, e.g. `AddCardEvent` and `AddCardFragment` etc, mainly to call out the origin of the event. Another option would be to name the event based on the action that should be taken. I'm up for changing this though in the long run.

sshropshire: Once all the activities are gone we should review the overall structure of DropIn. A big goal with the events is to decouple the host activity from the fragments and make each component easier to test.

Drop-In/src/main/java/com/braintreepayments/api/DropInPaymentMethodType.java
sshropshire: We've moved away from `Set` in our SDK interfaces a bit. There aren't any huge benefits with `Set` that we're taking advantage of. We still enforce uniqueness internally then convert the set into a list.

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
scannillo: 👍

scannillo: Should we leave a TODO?

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
scannillo: Ah - I missed that, thanks!

scannillo:
Looks good to me - just a few remaining questions left in the PR that need responses.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Going to add a TODO here. We may be able to create a single `AppEvent` type to simplify everything.

sshropshire:
> Should there be an `Edit` button in the Vault `Recent` section of the payment selection screen?

This is addressed in the `VaultManagerFragment` PR that @sarahkoop started. There's a typo in the `SettingsFragment` xml that causes the button not to show. It'll be restored in a separate PR.

sshropshire:
> I haven't looked at Android Drop In in a while so I forget ... BUT was the `Next` button the Add Card view always enabled? It looks like it always enabled, even if the card # is empty or invalid. Just curious here

This is a bit of a UX issue. We'll revist everything with a final round of polish to make sure we have parity, but I've seen a lot of places that the UX best practice is to leave the button enabled and show a dialog/validation error message when the form is invalid.

Drop-In/src/androidTest/java/com/braintreepayments/api/AddCardFragmentUITest.kt
sshropshire: Done!

demerino (approved):

PR #242, opened 2021-07-01
Convert VaultManagerActivity to VaultManagerFragment

sarahkoop:
### Summary of changes

 - Replace `VaultManagerActivity` with `VaultManagerFragment`
 
To be handled in future PRs:
- Testing DropInActivity (after all fragments have been extracted)
- Displaying loading indicator
- Testing analytics events

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire 
- @sarahkoop


Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -133,27 +143,80 @@ void onUIEvent(DropInUIEvent event) {
...
+                        removePaymentMethodNonce(paymentMethodNonceToDelete);
+                    }
+                })
+                .setOnDismissListener(new DialogInterface.OnDismissListener() {
+                    @Override
+                    public void onDismiss(DialogInterface dialog) {
+                        getDropInClient().sendAnalyticsEvent("manager.delete.confirmation.negative");
+                    }
+                })
+                .setNegativeButton(R.string.bt_cancel, new DialogInterface.OnClickListener() {
```
tdchow, line 176: I think you can pass `null` for the dialog click listener. It should function the same as setting an empty listener.

Drop-In/src/main/java/com/braintreepayments/api/VaultManagerFragment.java
```diff
@@ -0,0 +1,93 @@
...
+    @VisibleForTesting
+    DropInViewModel dropInViewModel;
+
+    public VaultManagerFragment() {
+    }
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+    }
```
tdchow, line 40: Can we omit this `onCreate` override since we're only calling the super?

tdchow (approved):
The changes are looking good! I highlighted a few clean-up items that shouldn't hold the PR.

Drop-In/src/androidTest/java/com/braintreepayments/api/VaultManagerFragmentUITest.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    @Test
+    fun whenStateIsRESUMED_onDeleteIconClick_sendsDeletePaymentEvent() {
+        val scenario = FragmentScenario.launchInContainer(VaultManagerFragment::class.java)
+        scenario.moveToState(Lifecycle.State.RESUMED)
+        scenario.onFragment { fragment ->
+            fragment.dropInViewModel.setVaultedPaymentMethods(vaultedPaymentMethodNonces)
+        }
```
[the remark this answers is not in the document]
sarahkoop: Yes - moved the repeated steps to beforeEach

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sarahkoop: Good call - changed this to null

Drop-In/src/main/java/com/braintreepayments/api/VaultManagerFragment.java
sarahkoop: Done 👍

PR #252, opened 2021-07-20
Handle 3DS Cancellation

sarahkoop:
### Summary of changes

 - Call `handleThreeDSecureActivityResult` and `deliverBrowserSwitchResult` from `DropInActivity` to handle 3DS errors and user cancellations
 - Use `dropInViewModel.isLoading` to handle animated button view loading switching and set `isLoading` false on user cancellations

 ### Checklist

### Authors

- @sarahkoop


Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
```diff
@@ -376,5 +376,6 @@ class CardDetailsFragmentUITest {
 
     fun whenStateIsRESUMED_whenThreeDSecureUserCancelationPresentInViewModel_showsSubmitButtonAgain() {
         // TODO: observe view model for UserCancelation cardTokenizationError and reset submit button
+        // TODO: is storing 3DS canceled in a view model necessary?
```
sarahkoop, line 379: We had a TODO for the approach of storing user canceled in view model but I think this can be done more simply by setting is loading to false in the view model from DropInActivity. Do you still think storing user canceled is necessary?

Drop-In/src/test/java/com/braintreepayments/api/DropInActivityUnitTest.java
```diff
@@ -85,6 +87,153 @@ public void onCreate_whenAuthorizationIsInvalid_finishesWithError() {
...
+
+        DropInClient dropInClient = new MockDropInClientBuilder()
+                .handleThreeDSecureActivityResultSuccess(result)
+                .build();
+
+        setupDropInActivity(authorization, dropInClient, dropInRequest, "sessionId");
+        mActivityController.setup();
+
+        assertFalse(mActivity.isFinishing());
+        // TODO assert finished with result
```
sarahkoop, line 121: How should we approach verifying finish with result is called (when `mActivity` is not a mock)?

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -44,10 +45,29 @@
...
+    protected void onResume() {
+        super.onResume();
+
+        getDropInClient().deliverBrowserSwitchResult(this, new DropInResultCallback() {
+            @Override
+            public void onResult(@Nullable DropInResult dropInResult, @Nullable Exception error) {
+                if (dropInResult != null) {
+                    finishWithDropInResult(dropInResult);
+                } else if (error instanceof UserCanceledException) {
+                    dropInViewModel.setIsLoading(false);
```
sarahkoop, line 63: Handling cancellation by setting `isLoading` to false in `dropInViewModel`. `CardDetailsFragment` is resumed on 3DS cancellation and observes the change in `isLoading` to set the animated button view to clickable again.

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sshropshire: This is almost UX territory. I feel like the current version finishes `DropInActivity` on user cancellation.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: I feel like our loading logic is going to evolve over time since there are a few more structural PRs we can do to clean up some of the event code. I wonder if it's something we can tackle in another PR since it's technically polish.
🇵🇱

Drop-In/src/test/java/com/braintreepayments/api/DropInActivityUnitTest.java
sshropshire: I think Robolectric might allow us to inspect the activity result we pass back to the Merchant. I forget exactly how.

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sarahkoop: I ran through the 3DS cancellation flow in the current version of Drop in and when a user cancels out of the 3DS screen they are returned to the card details screen (with the button clickable to submit again). That seems like the right flow to me, to make it easy for a user to re-try the same payment method, but do you think we should change that logic to Dismiss drop in if the user cancels?

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sarahkoop: This change was based on the assumption that the flow should work the same way as the current version, so we need to reset the loading status of the submit button when a user cancels.  I think that change makes sense in this PR if we are going with the previous version flow. If we want to dismiss Drop in all together on 3DS cancel, I will take out the loading logic.

Drop-In/src/test/java/com/braintreepayments/api/DropInActivityUnitTest.java
sarahkoop: Ah good call will look into that 👍

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sshropshire: Ah yeah I remember now. We should definitely do the same thing yeah. I think since `DropInActivity` presents the `CardDetailsFragment`, it has knowledge that it exists. So what we can do is `findFragmentByTag` and call `CardDetailsFragment#notifyTokenizationError()` or something like that to notify the user and reset the button.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Yeah we might be able to tackle this in another PR, but I'm starting to feel like there will be more than one loader as we get closer to GA. I think we're overloading the `DropInViewModel#isLoading` property a lot; it isn't clear which loader that value binds to at the moment.

scannillo: I agree with Steven that it is getting a little confusing.

> CardDetailsFragment is resumed on 3DS cancellation and observes the change in isLoading to set the animated button view to clickable again.

This isn't immediately obvious, without digging into CardDetailsFragment code. We could have an additional property (instead of re-using `isLoading`) with a more distinct purpose/name solely for signaling the submit button?

sarahkoop:
@scannillo @sshropshire Thanks for the initial feedback - I've updated this PR to set a userCanceledError in the view model and show the submit button again based on that error. This PR is now specifically to handle 3DS cancellation (we will need a separate PR to handle PayPal cancel flow). I've updated the tests but will hold off on full Activity testing until we can confirm our strategy there.

scannillo (approved):
🚀

sshropshire (approved):

PR #253, opened 2021-07-20
Refactor Events

sshropshire:
### Summary of changes

 - Refactor events emitted from `Fragment` results to use a single type: `DropInEvent`
 - The events decouple the fragments from their parent `Activity`, and allow for simpler UI testing
 methods

 ### Checklist

### Authors

- @sshropshire


Drop-In/src/main/java/com/braintreepayments/api/DropInEventType.java
```diff
@@ -0,0 +1,14 @@
+package com.braintreepayments.api;
+
+public enum DropInEventType {
+    ADD_CARD_SUBMIT,
+    CARD_DETAILS_SUBMIT,
+    DELETE_VAULTED_PAYMENT_METHOD,
+    DID_DISPLAY_SUPPORTED_PAYMENT_METHODS,
+    DISMISS_VAULT_MANAGER,
+    EDIT_CARD,
```
sarahkoop, line 9:
```suggestion
    EDIT_CARD_NUMBER,
```
This event is fired when a user clicks into the card number edit text to edit the number right? Want to clarify it.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -75,47 +70,58 @@ public void onResult(@Nullable List<DropInPaymentMethodType> paymentMethods, @Nu
         getSupportFragmentManager().setFragmentResultListener("BRAINTREE_EVENT", this, new FragmentResultListener() {
             @Override
             public void onFragmentResult(@NonNull String requestKey, @NonNull Bundle result) {
-                handleBraintreeEventBundle(result);
+                DropInEvent event = result.getParcelable("BRAINTREE_RESULT");
```
sarahkoop, line 73: Let's extract "BRAINTREE_RESULT" into a static String

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -200,8 +192,17 @@ public void onResult(@Nullable PaymentMethodNonce deletedNonce, @Nullable Except
...
     }
 
-    void sendAnalyticsEvent(String eventFragment) {
-        getDropInClient().sendAnalyticsEvent(eventFragment);
+    private void onSendAnalytics(DropInEvent event) {
+        String eventName = event.getString(DropInEventProperty.ANALYTICS_EVENT_NAME);
+        sendAnalyticsEvent(eventName);
+    }
+
+    void sendAnalyticsEvent(String eventName) {
```
sarahkoop, line 200: Can this be private now too?

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -405,6 +426,10 @@ public void finish() {
         overridePendingTransition(android.R.anim.fade_in, android.R.anim.fade_out);
     }
 
+    private void onVaultedPaymentMethodSelected(DropInEvent event) {
+        onVaultedPaymentMethodSelected(event.getPaymentMethodNonce(DropInEventProperty.VAULTED_PAYMENT_METHOD_SELECTION));
+    }
+
     public void onVaultedPaymentMethodSelected(final PaymentMethodNonce paymentMethodNonce) {
```
sarahkoop, line 433: Is this method called outside of DropInActivity? Can we make this private also and refactor the name? Not sure what name makes most sense here but it seems like we don't usually have overloaded method naming.

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,116 @@
...
+
+    static DropInEvent createVaultedPaymentMethodSelectedEvent(PaymentMethodNonce paymentMethodNonce) {
+        DropInEvent event = new DropInEvent(DropInEventType.VAULTED_PAYMENT_METHOD_SELECTED);
+        event.putParcelable(DropInEventProperty.VAULTED_PAYMENT_METHOD_SELECTION, paymentMethodNonce);
+        return event;
+    }
+
+    static DropInEvent createSupportedPaymentMethodSelectedEvent(DropInPaymentMethodType type) {
+        DropInEvent event = new DropInEvent(DropInEventType.SUPPORTED_PAYMENT_METHOD_SELECTED);
+        event.putString(DropInEventProperty.SUPPORTED_PAYMENT_METHOD_SELECTION, type.name());
```
sarahkoop, line 50: Nitpick - but SELECTION doesn't feel like the right naming here. How do you feel about `VAULTED_PAYMENT_METHOD_NONCE` and `SUPPORTED_PAYMENT_METHOD_TYPE`

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,116 @@
...
+
+    static DropInEvent createAddCardSubmitEvent(String cardNumber) {
+        DropInEvent event = new DropInEvent(DropInEventType.ADD_CARD_SUBMIT);
+        event.putString(DropInEventProperty.CARD_NUMBER, cardNumber);
+        return event;
+    }
+
+    static DropInEvent createCardDetailsSubmitEvent(Card card) {
+        DropInEvent event = new DropInEvent(DropInEventType.CARD_DETAILS_SUBMIT);
+        event.putParcelable(DropInEventProperty.CARD_DETAILS, card);
```
sarahkoop, line 26: Maybe we could just call this `CARD` instead of `CARD_DETAILS` to make it more clear that it's a card object being stored?

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,116 @@
...
+    DropInPaymentMethodType getDropInPaymentMethodType(DropInEventProperty property) {
+        String paymentMethodTypeString = payload.getString(property.getBundleKey());
+        return DropInPaymentMethodType.valueOf(paymentMethodTypeString);
+    }
+
+    PaymentMethodNonce getPaymentMethodNonce(DropInEventProperty property) {
+        return (PaymentMethodNonce) payload.getParcelable(property.getBundleKey());
+    }
+
+    public Card getCard(DropInEventProperty property) {
```
sarahkoop, line 86:
```suggestion
    Card getCard(DropInEventProperty property) {
```

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,116 @@
...
+        return (Card) payload.getParcelable(property.getBundleKey());
+    }
+
+    public DropInEventType getType() {
```
sarahkoop, line 90:
```suggestion
    DropInEventType getType() {
```

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,116 @@
...
+import android.os.Bundle;
+import android.os.Parcel;
+import android.os.Parcelable;
+
+class DropInEvent implements Parcelable {
+
+    private final DropInEventType type;
+    private final Bundle payload;
+
+    static DropInEvent createSendAnalytics(String eventName) {
```
sarahkoop, line 12:
```suggestion
    static DropInEvent createSendAnalyticsEvent(String eventName) {
```
Nitpick for consistency

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Good call! It may have been necessary in the past before we had events. We can/should make this private.

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
sshropshire: +1 on the renaming. We can review and rename.

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
sshropshire: Yeah I'm up for clarifying. I wasn't super sure on the names tbh.

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,107 @@
+package com.braintreepayments.api;
+
+import android.os.Bundle;
+import android.os.Parcel;
+import android.os.Parcelable;
+
+class DropInEvent {
```
scannillo, line 7: This looks awesome - much easier to read / understand than the separate event classes prior

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,107 @@
...
+        return event;
+    }
+
+    static DropInEvent createCardDetailsSubmitEvent(Card card) {
+        DropInEvent event = new DropInEvent(DropInEventType.CARD_DETAILS_SUBMIT);
+        event.putParcelable(DropInEventProperty.CARD, card);
+        return event;
+    }
+
+    static DropInEvent createEditCardEvent(String cardNumber) {
```
scannillo, line 31: Take it or leave it comment 
```suggestion
    static DropInEvent createEditCardNumberEvent(String cardNumber) {
```

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
```diff
@@ -0,0 +1,107 @@
...
+        return event;
+    }
+
+    static DropInEvent createVaultedPaymentMethodSelectedEvent(PaymentMethodNonce paymentMethodNonce) {
+        DropInEvent event = new DropInEvent(DropInEventType.VAULTED_PAYMENT_METHOD_SELECTED);
+        event.putParcelable(DropInEventProperty.VAULTED_PAYMENT_METHOD, paymentMethodNonce);
+        return event;
+    }
+
+    static DropInEvent createSupportedPaymentMethodSelectedEvent(DropInPaymentMethodType type) {
```
scannillo, line 49: Nitpick - do we need to specify that the payment method is `supported`. Don't we only show supported payment methods?

sarahkoop: In my opinion it's nice to differentiate between a vaulted payment method and a new (supported) payment method, but maybe there is a better name for those?

scannillo (approved):
Two small comments - otherwise, looks good!

sarahkoop (approved):

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
sshropshire: Yeah I agree in general. I'd be down to make these more concise if possible. The factory methods do make it easier when constructing an event though. It allows us to keep event creation logic in one place for the most part.

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
sshropshire: Def!

Drop-In/src/main/java/com/braintreepayments/api/DropInEvent.java
sshropshire: 🙏 ! It can still evolve too. The closer we get to GA I think we can clear up even more.

sshropshire:
@sarahkoop @scannillo merging this! I think we can continue to improve naming conventions in separate PRs as we move closer to GA.

PR #255, opened 2021-07-30
Handle Payment Cancellations

sarahkoop:
### Summary of changes

 - Handle Activity results for PayPal, GooglePay, and Venmo flows
 - Remove loading indicator and display supported payment methods in `SupportedPaymentMethodsFragment` on user cancellation 

 ### Checklist

- @sarahkoop

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
@@ -428,13 +428,7 @@ public class DropInActivity extends BaseActivity {
         getDropInClient().handleThreeDSecureActivityResult(this, resultCode, data, new DropInResultCallback() {
             @Override
             public void onResult(@Nullable DropInResult dropInResult, @Nullable Exception error) {
-                if (dropInResult != null) {
-                    finishWithDropInResult(dropInResult);
-                } else if (error instanceof UserCanceledException) {
-                    dropInViewModel.setUserCanceledError(error);
-                } else {
-                    onError(error);
-                }
+                onDropInResult(dropInResult, error);
             }
         });
     }
@@ -443,13 +437,7 @@ public class DropInActivity extends BaseActivity {
         getDropInClient().handleGooglePayActivityResult(this, resultCode, data, new DropInResultCallback() {
             @Override
             public void onResult(@Nullable DropInResult dropInResult, @Nullable Exception error) {
-                if (dropInResult != null) {
-                    finishWithDropInResult(dropInResult);
-                } else if (error instanceof UserCanceledException) {
-                    dropInViewModel.setUserCanceledError(error);
-                } else {
-                    onError(error);
-                }
+                onDropInResult(dropInResult, error);
             }
         });
     }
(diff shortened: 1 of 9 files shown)


Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -414,20 +408,53 @@ public void finish() {
...
+    private void handleVenmoActivityResult(int resultCode, Intent data) {
+        getDropInClient().handleVenmoActivityResult(this, resultCode, data, new DropInResultCallback() {
+            @Override
+            public void onResult(@Nullable DropInResult dropInResult, @Nullable Exception error) {
+                onDropInResult(dropInResult, error);
+            }
+        });
+    }
+
+    private void onDropInResult(DropInResult dropInResult, Exception error) {
```
sarahkoop, line 448: Not sure if this is the best name for this method. Open to other ideas here.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -414,20 +408,53 @@ public void finish() {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
+        switch (requestCode) {
```
sshropshire, line 411: Could we have a single `DropInClient#handleActivityResult()` that does this switch internally and forwards to the appropriate client based on request code?

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: I'm not opposed to this name it makes sense coming from a `DropInResultCallback` I feel.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sarahkoop: Good call - I moved it all into one `handleActivityResult` method that calls the `DropInClient` methods.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -414,18 +408,43 @@ public void finish() {
...
-                if (dropInResult != null) {
-                    finishWithDropInResult(dropInResult);
-                } else if (error instanceof UserCanceledException) {
-                    dropInViewModel.setUserCanceledError(error);
-                } else {
-                    onError(error);
-                }
-            }
-        });
+        handleActivityResult(requestCode, resultCode, data);
```
sshropshire, line 411: I think we can even take it a step further and call `getDropInClient()#handleActivityResult` here since we have access to the request codes from within `DropInClient`. Then `DropInActivity` only has to call one method and forward the result to the merchant.

sarahkoop: Ah got it ok moved this whole method to DropInClient 👍

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
scannillo: Agreed!

scannillo (approved):

sshropshire (approved):
🌊

PR #256, opened 2021-07-30
Back button

sshropshire:
### Summary of changes

 - Leverage `OnBackPressedCallback` to handle back button presses throughout DropIn
 - Embed Toolbars in each `Fragment` rather than in the top-level `DropInActivity`
 - Add an invisible button to allow the user to tap the transparent area to exit DropIn

 ### Checklist

### Authors

- @sshropshire

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/AddCardFragment.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/AddCardFragment.java
@@ -8,6 +8,7 @@ import android.view.ViewGroup;
 
 import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.VisibleForTesting;
+import androidx.appcompat.widget.Toolbar;
 import androidx.fragment.app.Fragment;
 import androidx.lifecycle.Observer;
 import androidx.lifecycle.ViewModelProvider;
@@ -26,6 +27,9 @@ public class AddCardFragment extends Fragment implements OnCardFormSubmitListene
 
     @VisibleForTesting
     CardForm cardForm;
+
+    private Toolbar toolbar;
+
     private SupportedCardTypesView supportedCardTypesView;
     private AnimatedButtonView animatedButtonView;
 
@@ -80,6 +84,14 @@ public class AddCardFragment extends Fragment implements OnCardFormSubmitListene
             public void handleOnBackPressed() {
                 getParentFragmentManager().popBackStack();
                 remove();
+            }
+        });
+
+        toolbar = view.findViewById(R.id.bt_toolbar);
+        toolbar.setNavigationOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                getParentFragmentManager().popBackStack();
             }
         });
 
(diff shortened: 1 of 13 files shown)


Drop-In/src/main/res/layout/bt_fragment_select_payment_method_parent.xml
```diff
@@ -16,6 +16,17 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
+    <Button
```
sarahkoop, line 19: I think for accessibility and to align with iOS instead of clicking outside to cancel we should add a cancel button. Here's an [iOS PR](https://github.com/braintree/braintree-ios-drop-in/pull/279) that shows the refactoring for this.

Drop-In/src/main/java/com/braintreepayments/api/SelectPaymentMethodParentFragment.java
```diff
@@ -65,37 +70,61 @@ public void transformPage(@NonNull View page, float position) {
...
+        });
+
+        Button backButton = view.findViewById(R.id.back_button);
+        backButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                sendDropInEvent(new DropInEvent(DropInEventType.CANCEL_DROPIN));
+            }
+        });
+
```
sarahkoop, line 96: Can we add some UI tests for this back button functionality / event dispatching?

Drop-In/src/main/java/com/braintreepayments/api/AddCardFragment.java
```diff
@@ -74,6 +77,22 @@ public void onChanged(Exception error) {
             }
         });
 
+        requireActivity().getOnBackPressedDispatcher().addCallback(getViewLifecycleOwner(), new OnBackPressedCallback(true) {
```
sarahkoop, line 80: I like this approach 👍

scannillo (approved):

Drop-In/src/main/java/com/braintreepayments/api/SelectPaymentMethodParentFragment.java
sshropshire: I was able to add a few back button tests. I can't figure out a way to test out the `popBackStack()` related tests right now though. I tried a good bit but it doesn't behave like expected.

Drop-In/src/main/res/layout/bt_fragment_select_payment_method_parent.xml
sshropshire: I think this may be Android UX/visual territory, maybe we can table for now? The screen reader will announce this as a button for now. From an accessibility standpoint we should be good.

sarahkoop: Yeah we can revisit that with additional UI changes later 👍

sarahkoop (approved):

PR #257, opened 2021-07-30
Animate Bottom Sheet

sshropshire:
### Summary of changes

 - Add bottom sheet transition animations to DropIn

https://user-images.githubusercontent.com/58225613/127711383-24c41882-dbed-4656-b6fa-b3eaefaf77cb.mov

 ### Checklist

### Authors

- @sshropshire

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/NoAnimationPageTransformer.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/NoAnimationPageTransformer.java
@@ -5,7 +5,7 @@ import android.view.View;
 import androidx.annotation.NonNull;
 import androidx.viewpager2.widget.ViewPager2;
 
-public class NoAnimationPageTransformer implements ViewPager2.PageTransformer {
+class NoAnimationPageTransformer implements ViewPager2.PageTransformer {
     @Override
     public void transformPage(@NonNull View page, float position) {
         if (position < -1 || position > 1) {
--- a/Drop-In/src/main/java/com/braintreepayments/api/NoAnimationPageTransformer.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/NoAnimationPageTransformer.java
@@ -1,0 +1,19 @@
+package com.braintreepayments.api;
+
+import android.view.View;
+
+import androidx.annotation.NonNull;
+import androidx.viewpager2.widget.ViewPager2;
+
+public class NoAnimationPageTransformer implements ViewPager2.PageTransformer {
+    @Override
+    public void transformPage(@NonNull View page, float position) {
+        if (position < -1 || position > 1) {
+            // page is either offscreen to the left or offscreen to the right
+            page.setAlpha(0.0f);
+        } else {
+            // page is visible
+            page.setAlpha(1.0f);
+        }
+    }
+}
--- a/Drop-In/src/main/java/com/braintreepayments/api/SelectPaymentMethodParentFragment.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/SelectPaymentMethodParentFragment.java
@@ -10,6 +10,7 @@ import android.view.ViewGroup;
 import android.view.animation.AccelerateInterpolator;
 import android.view.animation.DecelerateInterpolator;
 import android.widget.Button;
+import android.widget.ViewAnimator;
 
 import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.NonNull;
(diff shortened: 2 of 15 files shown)


scannillo:
It looks like the sheet enters with a slide in, but exits with a fade out. Would it be more consistent to exit with a slide out, as well?

Drop-In/src/main/java/com/braintreepayments/api/SelectPaymentMethodParentFragment.java
```diff
@@ -84,26 +114,46 @@ void onDropInEvent(DropInEvent event) {
     }
 
     private void onShowVaultManager(DropInEvent event) {
-        int targetHeight = dpToPixels(400);
-        ViewGroup.LayoutParams layoutParams = viewPager.getLayoutParams();
-        layoutParams.height = targetHeight;
-        viewPager.setLayoutParams(layoutParams);
+        // keep the same height when transitioning to vault manager
```
sarahkoop, line 117: The consistent height looks much better 👍

Drop-In/src/main/java/com/braintreepayments/api/SelectPaymentMethodParentFragment.java
```diff
@@ -68,6 +73,31 @@ public void transformPage(@NonNull View page, float position) {
         return view;
     }
 
+    @Override
+    public void onResume() {
+        super.onResume();
+        slideUpBottomSheet();
+    }
+
+    private void slideUpBottomSheet() {
```
sarahkoop, line 82: Should we add the same slide up animation on the vaulted cards? Those appear much more quickly than the bottom sheet slide up and might look better with the same animation.

sshropshire: Yeah I agree. I created a `ViewPager2Animator` class to make the animation more like the others. `ViewPager2` doesn't have an API to control smooth scroll speed so it has to be custom.

sshropshire:
> It looks like the sheet enters with a slide in, but exits with a fade out. Would it be more consistent to exit with a slide out, as well?

@scannillo I agree with this. It's kind of dependent on that back button PR. We could wait a bit until that get's merged and we can do something similar for a slide-down animation.

sshropshire:
@scannillo bottom sheet slide-out animation should be good to go!

Drop-In/src/main/res/layout/bt_fragment_add_card.xml
```diff
@@ -1,52 +1,39 @@
...
+    xmlns:bt="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:orientation="vertical"
+    android:background="@color/bt_base_background"
     tools:context="com.braintreepayments.api.AddCardFragment">
 
-    <androidx.appcompat.widget.Toolbar
-        android:id="@+id/bt_toolbar"
```
sarahkoop, line 10: I think we need this toolbar right? I'm getting an error in the demo app from AddCardFragment with this missing.

sshropshire: Yeah this was a rebase error i'm going to revert.

sshropshire: Great catch! It should be good now.

Drop-In/src/main/res/layout/bt_fragment_select_payment_method_parent.xml
```diff
@@ -10,6 +10,7 @@
         android:id="@+id/background"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
+        android:alpha="0"
```
sarahkoop, line 13: Not sure if this alpha is the cause or something else but the AddCardFragment is now transparent

sshropshire: Nice yeah that's my b. There were two things from that rebase that didn't make it. We have to explicitly set backgrounds on the individual fragments now since the activity is transparent.

Drop-In/src/main/java/com/braintreepayments/api/NoAnimationPageTransformer.java
```diff
@@ -0,0 +1,19 @@
+package com.braintreepayments.api;
+
+import android.view.View;
+
+import androidx.annotation.NonNull;
+import androidx.viewpager2.widget.ViewPager2;
+
+public class NoAnimationPageTransformer implements ViewPager2.PageTransformer {
```
sarahkoop, line 8: Can this be package private?

sarahkoop (approved):
The animations look good to me! One last nitpick - when cancelling PayPal / pressing back from the card number entry page the bottom sheet re-animates and slides up. I think it's ok as is but if we can not re-animate on cancellation that would probably look better.

scannillo (approved):

Drop-In/src/main/java/com/braintreepayments/api/NoAnimationPageTransformer.java
sshropshire: Yup!

sshropshire:
> The animations look good to me! One last nitpick - when cancelling PayPal / pressing back from the card number entry page the bottom sheet re-animates and slides up. I think it's ok as is but if we can not re-animate on cancellation that would probably look better.

Good call I saw this too! I also have an idea on how to resolve this that fits nicely under a saved instance state PR that we can make when we're doing some final polish. I have an idea about how we can extract some of that animation logic into a `BottomSheetPresenter` class too, I think those items could go hand in hand.

PR #263, opened 2021-08-04
Fix Demo App Integration Tests

sarahkoop:
### Summary of changes

 - Move the code for launching the app for integration testing out of the `setup` method so it can be called after any additional settings changes have been set.
 - Clean up TestHelper
 - Get all tests in `androidTest` in the Demo app passing

 ### Checklist

- @sarahkoop

### DIFF
--- a/Demo/src/androidTest/java/com/braintreepayments/demo/test/utilities/TestHelper.java
+++ b/Demo/src/androidTest/java/com/braintreepayments/demo/test/utilities/TestHelper.java
@@ -1,11 +1,7 @@
 package com.braintreepayments.demo.test.utilities;
 
-import android.Manifest;
 import android.content.Context;
-import android.content.pm.PackageManager;
-import android.os.SystemClock;
 import android.preference.PreferenceManager;
-import android.util.Log;
 import android.widget.Spinner;
 
 import androidx.annotation.CallSuper;
@@ -13,9 +9,6 @@ import androidx.test.core.app.ApplicationProvider;
 
 import com.braintreepayments.DeviceAutomator;
 import com.braintreepayments.cardform.view.CardForm;
-
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 
 import static com.braintreepayments.AutomatorAction.click;
 import static com.braintreepayments.AutomatorAction.setText;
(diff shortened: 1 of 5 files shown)


Drop-In/src/androidTest/java/com/braintreepayments/api/SelectPaymentMethodParentFragmentUITest.kt
```diff
@@ -40,6 +40,7 @@ class SelectPaymentMethodParentFragmentUITest {
 
     @Test
     fun whenStateIsRESUMED_onShowVaultManagerEvent_showsVaultManager() {
+        // TODO: determine how to test UI with animation delay
```
sarahkoop, line 43: This test is not in the Demo app integration tests so not specifically covered in this PR, but it's the only one still failing - I think due to the animation changes. The functionality is working as expected but the test fails because the assertion doesn't wait for the animation to complete. I couldn't find a good way to test with animation delay here. We could disable animation on the simulator before running, but I'd rather not add another manual step to the testing process. Maybe we should remove this test and test the functionality through a Demo app integration test?

Demo/src/androidTest/java/com/braintreepayments/demo/test/utilities/TestHelper.java
```diff
@@ -74,21 +63,13 @@ public void setMerchantAccountId(String merchantAccountId) {
...
-        onDevice(withText("Reset")).perform(click());
     }
 
     public void useTokenizationKey() {
         PreferenceManager.getDefaultSharedPreferences(ApplicationProvider.getApplicationContext())
                 .edit()
                 .putBoolean("tokenization_key", true)
                 .commit();
-
-        // additional sleep here makes tests more consistent for some reason
```
sarahkoop, line 88: The tests have been passing consistently without all these sleeps because we no longer need to call reset when we don't launch the app until after the settings have been configured. I think it's ok to remove all these now, but can add them back if the tests start to flake again.

Drop-In/src/androidTest/java/com/braintreepayments/api/SelectPaymentMethodParentFragmentUITest.kt
sshropshire: We could I ran across this issue in the refactor loader branch though. I put in a few wait for 1-second statements to get them all to pass.

sarahkoop: Ah good call - added a wait to this test 👍

Demo/src/androidTest/java/com/braintreepayments/demo/test/utilities/TestHelper.java
sshropshire: 🙌

sshropshire (approved):
🔥

scannillo (approved):

PR #264, opened 2021-08-05
Handle Errors in DropInActivity

sarahkoop:
### Summary of changes

 - Handle errors called back in payment flows in `DropInActivity`
 - Add unit tests for existing error handling

 ### Checklist

- @sarahkoop

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
@@ -378,7 +378,11 @@ public class DropInActivity extends BaseActivity {
         getDropInClient().getSupportedCardTypes(new GetSupportedCardTypesCallback() {
             @Override
             public void onResult(List<String> supportedCardTypes, Exception error) {
-                dropInViewModel.setSupportedCardTypes(Arrays.asList(DropInPaymentMethodType.getCardsTypes(supportedCardTypes)));
+                if (supportedCardTypes != null) {
+                    dropInViewModel.setSupportedCardTypes(Arrays.asList(DropInPaymentMethodType.getCardsTypes(supportedCardTypes)));
+                } else if (error != null) {
+                    onError(error);
+                }
             }
         });
 
--- a/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
@@ -193,7 +193,7 @@ public class DropInActivity extends BaseActivity {
                 .show();
     }
 
-    private void removePaymentMethodNonce(PaymentMethodNonce paymentMethodNonceToDelete) {
+    void removePaymentMethodNonce(PaymentMethodNonce paymentMethodNonceToDelete) {
         // proactively remove from view model
         dropInViewModel.removeVaultedPaymentMethodNonce(paymentMethodNonceToDelete);
 
@@ -209,7 +209,7 @@ public class DropInActivity extends BaseActivity {
                     //mLoadingViewSwitcher.setDisplayedChild(0);
                 } else {
                     getDropInClient().sendAnalyticsEvent("manager.unknown.failed");
-                    // TODO: determine how to handle unexpected error when deleting payment method (previously finished drop in)
+                    onError(error);
                 }
             }
         });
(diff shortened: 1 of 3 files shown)


Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -209,7 +209,7 @@ public void onResult(@Nullable PaymentMethodNonce deletedNonce, @Nullable Except
                     //mLoadingViewSwitcher.setDisplayedChild(0);
                 } else {
                     getDropInClient().sendAnalyticsEvent("manager.unknown.failed");
-                    // TODO: determine how to handle unexpected error when deleting payment method (previously finished drop in)
+                    onError(error);
```
sarahkoop, line 212: Decided to go with the same flow as the previous versions, seems good to give merchants information on the error and they can handle it how they choose.

sshropshire (approved):
🌊

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -193,7 +193,7 @@ public void onDismiss(DialogInterface dialog) {
                 .show();
     }
 
-    private void removePaymentMethodNonce(PaymentMethodNonce paymentMethodNonceToDelete) {
+    void removePaymentMethodNonce(PaymentMethodNonce paymentMethodNonceToDelete) {
```
sshropshire, line 196: random q: could this still be private?

sarahkoop: Yes -  I made it package-private to make testing easier. This method gets called from the confirmation click in the cancel alert that we build in `deleteVaultedPaymentMethod`. Seemed more straightforward to test this method directly than mock the alert dialog but could investigate that if we'd rather keep this private.

sshropshire: Ah nice yeah. That makes sense. We could throw in an `@VisibleForTesting` annotation so a lint error occurs if someone tries to call it externally.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -378,7 +379,11 @@ private void startAddCardFlow(@Nullable String cardNumber) {
         getDropInClient().getSupportedCardTypes(new GetSupportedCardTypesCallback() {
             @Override
             public void onResult(List<String> supportedCardTypes, Exception error) {
-                dropInViewModel.setSupportedCardTypes(Arrays.asList(DropInPaymentMethodType.getCardsTypes(supportedCardTypes)));
+                if (supportedCardTypes != null) {
+                    dropInViewModel.setSupportedCardTypes(Arrays.asList(DropInPaymentMethodType.getCardsTypes(supportedCardTypes)));
+                } else if (error != null) {
+                    onError(error);
+                }
```
scannillo, line 386: Nitpick - is it more standard / common practice to check the error case first?

Drop-In/src/test/java/com/braintreepayments/api/DropInActivityUnitTest.java
```diff
@@ -602,6 +602,139 @@ public void onSupportedPaymentMethodSelected_withTypeUnknown_showsAddCardFragmen
         assertNotNull(mActivity.getSupportFragmentManager().findFragmentByTag("ADD_CARD"));
     }
 
+    @Test
+    public void tokenizeCard_onError_whenErrorNotErrorWithResponse_finishesWithError() {
```
scannillo, line 606: Does this need an equivalent test for `whenError Is ErrorWithResponse`

scannillo (approved):
Two small comments - overall looks good!

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sarahkoop: Yeah we typically do error first - flipped this

Drop-In/src/test/java/com/braintreepayments/api/DropInActivityUnitTest.java
sarahkoop: Good call - the handling of the ErrorWithResponse is tested with `CardDetailsFragment`, but I added a test that the error is set in the `dropInViewModel` with it is `ErrorWithResponse`

PR #265, opened 2021-08-06
Make DropInActivity More DRY

sshropshire:
### Summary of changes

 - Remove `BaseActivity`
 - Create `AlertPresenter`
 - Create `DropInFragment` base class
 - Add static factory methods for all DropIn fragments

 ### Checklist

### Authors

- @sshropshire


Drop-In/src/main/java/com/braintreepayments/api/DialogInteraction.java
```diff
@@ -0,0 +1,6 @@
+package com.braintreepayments.api;
+
+public enum DialogInteraction {
```
sarahkoop, line 3: Can this be package private and the associated callback be package private?

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -269,13 +315,22 @@ private void onAddCardSubmit(DropInEvent event) {
         showCardDetailsFragment(cardNumber);
     }
 
-    private boolean shouldAddFragment(String tag) {
+    private boolean fragmentAlreadyExists(String tag) {
```
sarahkoop, line 318: This logic is a little confusing to me - it looks like we use this method to determine if we should show a fragment or not, but this method looks like it's checking if a fragment with this tag exists, not if it's currently visible. Is that right, or does findFragmentByTag only return a fragment if it's currently visible?

Drop-In/src/test/java/com/braintreepayments/api/BaseActivityUnitTest.java
```diff
@@ -1,106 +0,0 @@
...
-    @Test
-    public void getDropInClient_setsClientTokenPresentWhenAClientTokenIsNotPresent() {
-        Intent intent = new Intent();
-        intent.putExtra(DropInClient.EXTRA_CHECKOUT_REQUEST, new DropInRequest());
-        setup(intent);
-
-        mActivity.getDropInClient();
-
-        assertFalse(mActivity.mClientTokenPresent);
-    }
```
sarahkoop, line 61: Should these be tested for DropInActivity now too?

sshropshire: When I was removing `BaseActivity` I couldn't see the value in these tests. We should have enough coverage in `DropInActivityUnitTest`.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Yeah visibility isn't a concern here, we just don't want to re-create a fragment if one already exists and is managed by the fragment manager. Guards us against those edge cases like a user double-tapping a button; we want to make sure we don't mistakenly put two Card Details fragments on the screen at once.

Drop-In/src/main/java/com/braintreepayments/api/DialogInteraction.java
sshropshire: Done!

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Went ahead and reverted this to `shouldAddFragment` that does read better.

Drop-In/src/main/java/com/braintreepayments/api/DialogInteractionCallback.java
```diff
@@ -0,0 +1,6 @@
+package com.braintreepayments.api;
+
+public interface DialogInteractionCallback {
```
sarahkoop, line 3: Can this be package private too?

PR #266, opened 2021-08-11
Fix PayPal Loader Bug

sshropshire:
### Summary of changes

 - Bump Core SDK Version to `4.4.0`

 ### Checklist

### Authors

- @sshropshire

### DIFF
--- a/Drop-In/build.gradle
+++ b/Drop-In/build.gradle
@@ -65,14 +65,14 @@ dependencies {
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'androidx.fragment:fragment:1.3.5'
 
-    api 'com.braintreepayments.api:braintree-core:4.4.0'
-    api 'com.braintreepayments.api:three-d-secure:4.4.0'
-    api 'com.braintreepayments.api:paypal:4.4.0'
-    api 'com.braintreepayments.api:venmo:4.4.0'
-    api 'com.braintreepayments.api:google-pay:4.4.0'
-    api 'com.braintreepayments.api:card:4.4.0'
-    api 'com.braintreepayments.api:data-collector:4.4.0'
-    api 'com.braintreepayments.api:union-pay:4.4.0'
+    api 'com.braintreepayments.api:braintree-core:4.4.1'
+    api 'com.braintreepayments.api:three-d-secure:4.4.1'
+    api 'com.braintreepayments.api:paypal:4.4.1'
+    api 'com.braintreepayments.api:venmo:4.4.1'
+    api 'com.braintreepayments.api:google-pay:4.4.1'
+    api 'com.braintreepayments.api:card:4.4.1'
+    api 'com.braintreepayments.api:data-collector:4.4.1'
+    api 'com.braintreepayments.api:union-pay:4.4.1'
 
     api 'com.braintreepayments:card-form:5.2.0'
 
--- a/Drop-In/build.gradle
+++ b/Drop-In/build.gradle
@@ -65,14 +65,14 @@ dependencies {
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'androidx.fragment:fragment:1.3.5'
 
-    api 'com.braintreepayments.api:braintree-core:4.1.0'
-    api 'com.braintreepayments.api:three-d-secure:4.1.0'
-    api 'com.braintreepayments.api:paypal:4.1.0'
-    api 'com.braintreepayments.api:venmo:4.1.0'
-    api 'com.braintreepayments.api:google-pay:4.1.0'
-    api 'com.braintreepayments.api:card:4.1.0'
-    api 'com.braintreepayments.api:data-collector:4.1.0'
-    api 'com.braintreepayments.api:union-pay:4.1.0'
+    api 'com.braintreepayments.api:braintree-core:4.4.0'
+    api 'com.braintreepayments.api:three-d-secure:4.4.0'
+    api 'com.braintreepayments.api:paypal:4.4.0'
+    api 'com.braintreepayments.api:venmo:4.4.0'
+    api 'com.braintreepayments.api:google-pay:4.4.0'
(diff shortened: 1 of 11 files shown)


Drop-In/build.gradle
```diff
@@ -65,14 +65,14 @@ dependencies {
...
 
-    api 'com.braintreepayments.api:braintree-core:4.1.0'
-    api 'com.braintreepayments.api:three-d-secure:4.1.0'
-    api 'com.braintreepayments.api:paypal:4.1.0'
-    api 'com.braintreepayments.api:venmo:4.1.0'
-    api 'com.braintreepayments.api:google-pay:4.1.0'
-    api 'com.braintreepayments.api:card:4.1.0'
-    api 'com.braintreepayments.api:data-collector:4.1.0'
-    api 'com.braintreepayments.api:union-pay:4.1.0'
+    api 'com.braintreepayments.api:braintree-core:4.4.0'
```
sarahkoop, line 68: Let's call out that we use core v4.2.0 modules in the CHANGELOG

Drop-In/src/main/java/com/braintreepayments/api/SupportedPaymentMethodsFragment.java
```diff
@@ -22,6 +23,12 @@
 
 public class SupportedPaymentMethodsFragment extends Fragment implements SupportedPaymentMethodSelectedListener, VaultedPaymentMethodSelectedListener {
 
+    private enum ViewState {
```
sarahkoop, line 26: Can I get a little more info on what this ViewState is needed for? I feel like throughout the SDK we have avoided maintaining a state like this

Drop-In/src/test/java/com/braintreepayments/api/DropInActivityUnitTest.java
```diff
@@ -778,4 +795,15 @@ private void assertExceptionIsReturned(String analyticsEvent, Exception exceptio
...
     }
+
+    private BrowserSwitchResult createBrowserSwitchSuccessResult() {
+        int requestCode = 123;
+        Uri url = Uri.parse("https://example.com");
+        JSONObject metadata = new JSONObject();
+        String returnUrlScheme = "sample-scheme";
+
+        BrowserSwitchRequest browserSwitchRequest =
+            new BrowserSwitchRequest(requestCode, url, metadata, returnUrlScheme, true);
```
sarahkoop, line 806:
```suggestion
            new BrowserSwitchRequest(requestCode, url, metadata, returnUrlScheme);
```
I don't remember what that true parameter was for, but looks like that method was changed in browser switch

Drop-In/src/test/java/com/braintreepayments/api/DropInClientUnitTest.java
```diff
@@ -1840,4 +1855,13 @@ private Configuration mockConfiguration(boolean paypalEnabled, boolean venmoEnab
 
         return configuration;
     }
+
+    private static BrowserSwitchResult createSuccessfulBrowserSwitchResult() {
+        int requestCode = 123;
+        Uri url = Uri.parse("www.example.com");
+        String returnUrlScheme = "sample-scheme";
+        BrowserSwitchRequest browserSwitchRequest = new BrowserSwitchRequest(
+                requestCode, url, new JSONObject(), returnUrlScheme, true);
```
sarahkoop, line 1864:
```suggestion
                requestCode, url, new JSONObject(), returnUrlScheme);
```

Same thing here - I think there may be a few things off in these tests after rebasing. They don't compile locally for me

Drop-In/src/test/java/com/braintreepayments/api/DropInRequestUnitTest.java
```diff
@@ -140,6 +140,7 @@ public void isParcelable() {
         billingAddress.setCountryCodeAlpha2("US");
         billingAddress.setPhoneNumber("3125557890");
         threeDSecureRequest.setBillingAddress(billingAddress);
+        threeDSecureRequest.setCardAddChallengeRequested(false);
```
sarahkoop, line 143: Is this needed? I thought this param was optional

Drop-In/src/main/java/com/braintreepayments/api/SupportedPaymentMethodsFragment.java
sshropshire: The main reason is to keep the view presentation logic in one place. Calling `setViewState` triggers a re-layout.

sshropshire: Honestly thinking about leveraging the presenter pattern here to separate concerns. This view is complex because the loader being `visible` is driven by a lot of factors external to this view.

Drop-In/src/test/java/com/braintreepayments/api/DropInRequestUnitTest.java
sshropshire: Yeah this was the bug that got fixed [#429](https://github.com/braintree/braintree_android/pull/429). It was throwing a null pointer exception.

Drop-In/build.gradle
sshropshire: Done!

Drop-In/src/main/java/com/braintreepayments/api/SupportedPaymentMethodsFragment.java
sshropshire: On second thought, [this](https://github.com/braintree/braintree-android-drop-in/blob/4be12a3b3ad1bce680c52b8245bf5d811d4e6b5a/Drop-In/src/main/java/com/braintreepayments/api/SupportedPaymentMethodsFragment.java#L138) was the primary reason. When this view comes to the foreground we have to decide whether or not to show the loader. When DropIn is `FINISHING` we allow the loader to remain on the screen.

PR #267, opened 2021-08-12
Dropin Finish Animation

sshropshire:
### Summary of changes

 - WIP
 - This PR adds back bottom sheet slide out animation when the activity is finished.

 ### Checklist

 - [ ] Added a changelog entry

### Authors

- @sshropshire


Drop-In/src/main/java/com/braintreepayments/api/BottomSheetFragment.java
```diff
@@ -123,23 +138,37 @@ public void onDestroy() {
...
+    private void slideDownBottomSheet() {
+        slideDownBottomSheet(null);
     }
 
-    private void cancelDropIn(@Nullable final AnimationCompleteCallback callback) {
-        if (bottomSheetPresenter.isAnimatingBottomSheet()) {
-            // prevent drop in cancellation while bottom sheet is animating
+    private void slideDownBottomSheet(@Nullable final AnimationCompleteCallback callback) {
+        if (bottomSheetPresenter.isAnimating()) {
+            // prevent drop in from being hidden while bottom sheet is animating
```
scannillo, line 161: These comments are helpful 👍

demerino (approved):
Video helps, thank you. Any need to vet this on a variety of devices/screen sizes? What about landscape?

scannillo:
I'm testing this on a Pixel 3a XL Emulator and am seeing this black bar sort of drop-down every time I dismiss the DropIn. It also shows a black bar whenever I open the DropIn sheet, but it is less noticeable then. Did we do any device testing, or could this be an emulator bug?

I can't replicate this on a Pixel 2 XL emulator. (At the same time, using that Emulator doesn't show the same smooth animations that your video does - everything is more jumpy).

I'm leaning to think that this only happening on my emulators, but wanted to check!

https://user-images.githubusercontent.com/35243507/129915126-5f3007ce-c283-4720-ad82-329364ebb0fa.mov

sshropshire:
> Video helps, thank you. Any need to vet this on a variety of devices/screen sizes? What about landscape?

Good call for the most part it should be good on landscape. But we should test more variations for sure. Probably a good idea to compare it against the current GA and see if there are any differences. Nothing should bug out though, since Android handles the majority of device rotation logic.

sshropshire:
> I'm testing this on a Pixel 3a XL Emulator and am seeing this black bar sort of drop-down every time I dismiss the DropIn. It also shows a black bar whenever I open the DropIn sheet, but it is less noticeable then. Did we do any device testing, or could this be an emulator bug?

Ah yeah I've seen this. I took a piece of code out because it seems like the Galaxy S9 has a different animation style. We basically need to disable Android's default activity animation.

sshropshire:
> I can't replicate this on a Pixel 2 XL emulator. (At the same time, using that Emulator doesn't show the same smooth animations that your video does - everything is more jumpy).

Also I realized a few weeks ago Debug builds are jumpy in nature. [Changing the build variant](https://developer.android.com/studio/run#changing-variant) to release will strip the debug symbols from the apk, and the animations should be smooth.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -305,6 +300,33 @@ public void onResult(@Nullable List<DropInPaymentMethodType> paymentMethods, @Nu
...
+        } else {
+            // assume drop in cancelled
+            sendAnalyticsEvent("sdk.exit.canceled");
+            setResult(RESULT_CANCELED);
+        }
+
+        if (cancelPendingActivityAnimation) {
+            // prevent default Android Activity animation from running
+            overridePendingTransition(0, 0);
+        }
```
sshropshire, line 326: @scannillo this should prevent the status bar at the top from animating. In some cases, we may want to keep the default animation though e.g. when the card entry form is showing and covers the entire screen, a fade-out/pop animation should look better than abruptly transitioning back to the merchant app.

\+ each Android phone manufacturer adds its own animations styles, so it'll look different on other devices / emulators.

scannillo: 👋  I just pulled and re-ran. On both the debug and release variant, I still see the black bar animating on the Pixel 3a XL. Is this also what you expect?

sarahkoop:
I'm also noticing that it looks like the toolbar is visible on the slide up animation (it's less obvious than the slide down bug Sammy is showing but you can also see this in the video she included). Are we ok with this?

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Let me double-check on an emulator I've been using a device. There there is one other piece in `styles.xml` I took out that may also be necessary.

sshropshire:
> I'm also noticing that it looks like the toolbar is visible on the slide up animation (it's less obvious than the slide down bug Sammy is showing but you can also see this in the video she included). Are we ok with this?

@sarahkoop @scannillo 🙏 &nbsp;🙏 &nbsp;the `@null` animation style in `styles.xml` is actually needed and [apparently](https://stackoverflow.com/a/20745135) `overridePendingTransition()` needs to be called after `finish()` 😬

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -382,13 +406,19 @@ public void onError(final Exception error) {
...
-        finish(dropInResult.getPaymentMethodNonce(), dropInResult.getDeviceData());
+    private void animateBottomSheetClosedAndFinishDropInWithResult(DropInResult dropInResult) {
+        pendingDropInResult = dropInResult;
+
+        if (isBottomSheetVisible()) {
+            // when the bottom sheet transitions to the "HIDDEN" state; the activity will finish
+            dropInViewModel.setBottomSheetState(BottomSheetState.HIDE_REQUESTED);
+        } else {
+            // no need to animate bottom sheet hidden; finish activity with default Android animation
+            finishDropInWithPendingResult(false);
```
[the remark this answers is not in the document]
sshropshire: Yeah when we tokenize a card, there is no bottom sheet to animate. The thought was to leverage the default activity `finish()` animation in this case.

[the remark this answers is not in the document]
sshropshire: Is there an issue with PayPal in this case? The logic could be off, but the bottom sheet should animate in that scenario.

sarahkoop:
https://user-images.githubusercontent.com/67924275/130825355-d2bc0bb0-9290-411a-aa71-0851d2316917.mov

PR #275, opened 2021-09-14
Signature Verification Fix v4

sarahkoop:
### Summary of changes

- Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
- Bump braintree_android version to 3.x.x - (version to be updated after v3 core release)

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop


build.gradle
```diff
@@ -24,5 +24,6 @@ allprojects {
         maven {
             url 'file:///tmp/maven/'
         }
+        mavenLocal()
```
sarahkoop, line 27: This is here for testing against a local release - will be removed after v3 core release

Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityUnitTest.java
```diff
@@ -975,10 +975,6 @@ public void showsSubmitButtonAgainWhenThreeDSecureIsCanceled() throws PackageMan
         setText(mEditCardView, R.id.bt_card_form_expiration, ExpirationDate.VALID_EXPIRATION);
         mEditCardView.findViewById(R.id.bt_button).performClick();
 
-        Intent nextStartedActivity = shadowOf(mActivity).peekNextStartedActivity();
```
sarahkoop, line 978: This `showsSubmitButtonAgainWhenThreeDSecureIsCanceled` unit test was failing after upgrading because of a null pointer on the `nextStartedActivity` here. I think it has something to do with testing library version and the shadow activity, but I couldn't pinpoint the issue. I'm not entirely sure what the value of these assertions are, and the test passed without them. I manually tested the 3DS cancel flow and verified that the button is visible. Are we ok with removing these two assertions?

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.0'
+        classpath 'com.android.tools.build:gradle:3.3.3'
```
sarahkoop, line 8: Android 11 requires upgrading gradle build tools to this minor version (or 4.0.1+). We can't upgrade to the latest 3.x core version without doing this, but are we ok with this as a non-breaking change? I think merchants will have to upgrade their build tools to the latest minor version also

Drop-In/src/test/java/com/braintreepayments/api/dropin/AddCardActivityUnitTest.java
sshropshire: Yeah these assertions seem unrelated to the actual named test. I'd be cool with removing those too since we really want to know that the user can retry after a 3DS cancellation.

sshropshire (approved):
🌊

Demo/build.gradle
```diff
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 28
-    buildToolsVersion '28.0.3'
+    compileSdkVersion 31
+    buildToolsVersion '30.0.3'
```
scannillo, line 5: Should buildToolsVersion also be 31?

Demo/src/androidTest/java/com/braintreepayments/demo/test/CardholderNameDropInTest.java
```diff
@@ -28,7 +28,7 @@ public void setup() {
     @Test(timeout = 60000)
     public void cardholderName_whenDisabled_isHidden() {
         setCardholderNameStatus(CardForm.FIELD_DISABLED);
-        onDevice(withText("Add Payment Method")).waitForExists().waitForEnabled().perform(click());
+        onDevice(withText("Add Payment Method")).waitForExists(2000).waitForEnabled(2000).perform(click());
```
scannillo, line 31: Is this in milliseconds? If not, 2000 is long lol

Demo/src/debug/AndroidManifest.xml
```diff
@@ -12,7 +12,8 @@
         <meta-data android:name="com.google.android.gms.wallet.api.enabled" android:value="true" />
 
         <activity
-            android:name=".MainActivity">
+            android:name=".MainActivity"
+            android:exported="true">
```
scannillo, line 16: Merchants need to add this to their apps right for Android 12 support? Did we call it out in all the right places? (Changelog, README, Docs)?

build.gradle
scannillo: I think we should call it out in the CHANGELOG or elsewhere if this BT DropIn update requires a tooling update on their end.

Demo/src/androidTest/java/com/braintreepayments/demo/test/CardholderNameDropInTest.java
sarahkoop: Yes this is two seconds lol, we made this change in later versions - seems to help stabilize the tests

Demo/src/debug/AndroidManifest.xml
sarahkoop: So merchants don't actually need to make this change until they update their own apps to target/compile for API 31 (Android 12). If they use this version in an app that doesn't target Android 12, the exported tags are still optional. They would need to add the exported flags when they update their target/compile version, but I think that's probably outside the scope of our CHANGELOG right? We did update our [docs examples](https://developer.paypal.com/braintree/docs/guides/client-sdk/setup/android/v4#browser-switch-setup) to include this though.

Demo/build.gradle
sarahkoop: Ah that's a good callout - I used the Android Studio auto-upgrade feature and it put us at 30.0.3 as the latest. After looking more into it, it looks like there is a version 31 that's been released, but they should be forwards compatible (if you don't specify buildToolsVersion the default min compatible version will be used). Anyways, these build tools are used for building, debugging Android apps, command line tools, and connecting to devices, so I don't think keeping it at this version should have a merchant impact.  We can look at upgrading all the SDKs once Android 12 is GA, but for now I'd opt to keep this as is since the other SDKs have been released with this tools version.

demerino (approved):

PR #276, opened 2021-09-17
Fix Integration Tests

sarahkoop:
### Summary of changes

 - Downgrade constraint layout dependency because it was causing errors since we don't explicitly require Java 8
 - Remove mavenLocal()
 - Align buildToolsVerions

 ### Checklist

- @sarahkoop

### DIFF
--- a/Demo/build.gradle
+++ b/Demo/build.gradle
@@ -58,11 +58,6 @@ android {
         exclude 'META-INF/maven/com.google.guava/guava/pom.properties'
         exclude 'META-INF/maven/com.google.guava/guava/pom.xml'
     }
-
-    compileOptions {
-        sourceCompatibility 1.8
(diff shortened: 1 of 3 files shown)


sshropshire (approved):

jplukarski (approved):
Just curious, do we have plans to explicitly require Java 8 so we can eventually update ConstraintLayout?

sarahkoop:
> Just curious, do we have plans to explicitly require Java 8 so we can eventually update ConstraintLayout?

Good question, Steven and I were discussing this - we have wanted to support Java 7 projects in the past, but we may want to explicitly require Java 8 in future major versions. It would probably make sense to make that requirement explicit in core before drop-in, though, and we still need to do some investigating here.

PR #277, opened 2021-09-17
Upgrade Robolectric to 4.6.1

sshropshire:
### Summary of changes

 - Upgrade Robolectric to version `4.6.1`
 - Fixup some tests that were failing after update

 ### Checklist

### Authors

- @sshropshire

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,10 +7,10 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
-      - name: Set up Java 8
+      - name: Set up Java 11
         uses: actions/setup-java@v2
         with:
-          java-version: '8'
+          java-version: '11'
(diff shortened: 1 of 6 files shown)


sarahkoop:
Do we need to bump to Java 11 in GitHub Actions for this change?

sshropshire:
> Do we need to bump to Java 11 in GitHub Actions for this change?

Yeah good call. I can bump.

sarahkoop:
Would it make sense to set compileOptions to Java 8 (or even Java 7?) in the demo app to make sure we don't introduce any incompatible changes in the SDK?

sshropshire:
> Would it make sense to set compileOptions to Java 8 (or even Java 7?) in the demo app to make sure we don't introduce any incompatible changes in the SDK?

Decided to go with explicit Java 7 support until we can investigate to see if browser switch creates an implicit Java 8 requirement.

PR #279, opened 2021-09-23
Fix Cardholder Name Field Focus

### LINKED ISSUE #180: Credit Card Payment asks for Number first and then other fields appear
Dear support, 

I have managed to successfully implement CC payment with Drop-In. but my problem with Credit card payment is:

first, it asks for only credit card number, after punching in the numbers it shows other fields with Expiry date field as selected. The Cardholder field appears late in the process and is mandatory. I would like the whole form to appear at once at start.

I have added two screenshots to explain.  Thanks for your support.

first step
second step

best
M

sarahkoop:
### Summary of changes

 - Focus Cardholder Name field when adding card details if field not disabled (fixes #180)
 - Currently the expiration date field is focused first when adding card details, even if the cardholder name field is displayed (i.e. not disabled). This PR updates the logic to focus the expiration date field only if cardholder name is disabled, otherwise focus cardholder name first. 

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
+++ b/Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
@@ -37,15 +37,45 @@ class CardDetailsFragmentUITest {
     }
 
     @Test
-    fun whenStateIsRESUMED_expirationDateFieldIsFocused() {
+    fun whenStateIsRESUMED_andCardholderNameDisabled_expirationDateFieldIsFocused() {
+        val dropInRequest = DropInRequest()
         val args = Bundle()
-        args.putParcelable("EXTRA_DROP_IN_REQUEST", DropInRequest())
+        args.putParcelable("EXTRA_DROP_IN_REQUEST", dropInRequest)
         args.putParcelable("EXTRA_CARD_FORM_CONFIGURATION", CardFormConfiguration(false, false))
         args.putString("EXTRA_CARD_NUMBER", VISA)
         val scenario = FragmentScenario.launchInContainer(CardDetailsFragment::class.java, args, R.style.bt_drop_in_activity_theme)
         scenario.moveToState(Lifecycle.State.RESUMED)
 
+        assertEquals(CardForm.FIELD_DISABLED, dropInRequest.cardholderNameStatus)
         onView(withId(R.id.bt_card_form_expiration)).check(matches(isFocused()))
+    }
+
+    @Test
+    fun whenStateIsRESUMED_andCardholderNameOptional_cardholderNameFieldIsFocused() {
+        val dropInRequest = DropInRequest()
+        dropInRequest.cardholderNameStatus = FIELD_OPTIONAL
+        val args = Bundle()
+        args.putParcelable("EXTRA_DROP_IN_REQUEST", dropInRequest)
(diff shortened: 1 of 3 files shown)


Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
```diff
@@ -37,17 +37,47 @@ class CardDetailsFragmentUITest {
...
+        val dropInRequest = DropInRequest()
         val args = Bundle()
-        args.putParcelable("EXTRA_DROP_IN_REQUEST", DropInRequest())
+        args.putParcelable("EXTRA_DROP_IN_REQUEST", dropInRequest)
         args.putParcelable("EXTRA_CARD_FORM_CONFIGURATION", CardFormConfiguration(false, false))
         args.putString("EXTRA_CARD_NUMBER", VISA)
         val scenario = FragmentScenario.launchInContainer(CardDetailsFragment::class.java, args, R.style.bt_drop_in_activity_theme)
         scenario.moveToState(Lifecycle.State.RESUMED)
 
+        assertEquals(CardForm.FIELD_DISABLED, dropInRequest.cardholderNameStatus)
```
sshropshire, line 49: nit: It feels like the `DropInRequest` constructor should be responsible for setting the default `FIELD_DISABLED` value.

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
```diff
@@ -37,17 +37,47 @@ class CardDetailsFragmentUITest {
     }
 
     @Test
-    fun whenStateIsRESUMED_expirationDateFieldIsFocused() {
+    fun whenStateIsRESUMED_andCardholderNameDisabled_expirationDateFieldIsFocused() {
+        val dropInRequest = DropInRequest()
```
sshropshire, line 41:
```suggestion
        val dropInRequest = DropInRequest()
        dropInRequest.cardholderNameStatus = FIELD_DISABLED
```

sshropshire (approved):
🌊 &nbsp; couple comments but overall LGTM.

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sarahkoop: Yeah the constructor does set it to FIELD_DISABLED by default, just wanted to make that explicit in the test, but I can remove the assertion since it is the default 👍

sarahkoop, line 49:
```suggestion
```

Drop-In/src/androidTest/java/com/braintreepayments/api/CardDetailsFragmentUITest.kt
sarahkoop: This is set to disabled by default

sshropshire: Right yeah. If that changes though this test could break.

PR #281, opened 2021-09-29
Compile with Java 8 to Fix Tests

sarahkoop:
### Summary of changes

 - Integration tests are currently failing with manifest merger errors (fixed by upgrading `androidx.text.ext.junit` version)
 - When junit is upgraded, tests fail with error solved [here](https://stackoverflow.com/questions/49891730/invoke-customs-are-only-supported-starting-with-android-0-min-api-26) by specifying Java 8 as compile version (instead of Java 7)
 - Unit tests fail in command line because of lint error - resolved by upgrading appcompat dependency to `1.4.0-alpha03`

**Note**: I tried to create a test Android app compiling with Java 7 and including drop-in as a dependency, and it worked before and after the change from compileOptions Java 7 to Java 8 in this PR - I can't confirm that nothing will break depending on gradle versions, but since this is a major version update it seems like a good time to make this change to ensure we don't have future conflicts as we try to use Java 11 for testing and other updated libraries. 

 ### Checklist

- @sarahkoop

### DIFF
--- a/Demo/build.gradle
+++ b/Demo/build.gradle
@@ -61,7 +61,7 @@ android {
 }
 
 dependencies {
-    implementation 'androidx.appcompat:appcompat:1.3.1'
+    implementation 'androidx.appcompat:appcompat:1.4.0-alpha03'
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
 
--- a/Drop-In/build.gradle
+++ b/Drop-In/build.gradle
@@ -30,8 +30,8 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_7.toString()
-        targetCompatibility JavaVersion.VERSION_1_7.toString()
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
     }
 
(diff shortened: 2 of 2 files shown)


sarahkoop:
After more investigation - downgrading the kotlin version to `1.4.10` resolves the integration test error when `compileOptions` are set to Java 7. 

Requiring Java 8 might still make sense for this major version to give us more flexibility in the future, but interested in other opinions here.

sshropshire:
> After more investigation - downgrading the kotlin version to `1.4.10` resolves the integration test error when `compileOptions` are set to Java 7.
> 
> Requiring Java 8 might still make sense for this major version to give us more flexibility in the future, but interested in other opinions here.

This is tough I'd like to keep supporting Java 7 if possible, but it is becoming a bit of a chore. And with Android Gradle Plugin implicitly choosing Java 8 for new apps, I feel like apps targeting Java 7 source compatability will be lower in the future.

I honestly feel like we declare a Java 8 target for now and re-evaluate later. My hope is that we can revert to Java 7 before GA. The UITests are breaking the build, and my hope is that we'll be able to migrate all of those to Robolectric. Once that happens `androidTest` for `:DropIn` will no longer require a Kotlin dependency.

sshropshire (approved):

jplukarski (approved):
LGTM

PR #293, opened 2021-10-27
Set `sessionId` and `integrationType` in `DropInClient`

sarahkoop:
### Summary of changes

 - Set `sessionId` and `integrationType` in `DropInClient` (this allows for the same session ID to be used for analytics events throughout the Drop-in flow)

 ### Checklist

- @sarahkoop

### DIFF
--- a/Drop-In/src/test/java/com/braintreepayments/api/DropInClientUnitTest.java
+++ b/Drop-In/src/test/java/com/braintreepayments/api/DropInClientUnitTest.java
@@ -65,9 +65,15 @@ public class DropInClientUnitTest {
     }
 
     @Test
-    public void constructor_setsBraintreeClientWithSessionId() {
+    public void publicConstructor_setsBraintreeClientWithSessionId() {
         DropInClient sut = new DropInClient(ApplicationProvider.getApplicationContext(), Fixtures.TOKENIZATION_KEY, new DropInRequest());
         assertNotNull(sut.braintreeClient.getSessionId());
+    }
+
+    @Test
+    public void internalConstructor_setsBraintreeClientWithSessionId() {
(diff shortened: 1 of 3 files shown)


sshropshire (approved):

Drop-In/src/test/java/com/braintreepayments/api/DropInClientUnitTest.java
```diff
@@ -56,6 +58,18 @@ public void beforeEach() {
...
+    @Test
+    public void constructor_setsIntegrationTypeDropIn() {
+        DropInClient sut = new DropInClient(ApplicationProvider.getApplicationContext(), Fixtures.TOKENIZATION_KEY, new DropInRequest());
+        assertEquals(IntegrationType.DROP_IN, sut.braintreeClient.getIntegrationType());
+    }
+
+    @Test
+    public void constructor_setsBraintreeClientWithSessionId() {
+        DropInClient sut = new DropInClient(ApplicationProvider.getApplicationContext(), Fixtures.TOKENIZATION_KEY, new DropInRequest());
+        assertNotNull(sut.braintreeClient.getSessionId());
```
scannillo, line 70: Can we do better than a `notNull` assert? I assume no, since we'd have to mock the analytics layer which is in `core`. Just wanted to poke on it though

scannillo (approved):

Drop-In/src/test/java/com/braintreepayments/api/DropInClientUnitTest.java
jplukarski: The public constructor generates a random ID. We added an additional test for the internal constructor that can take a session ID as an argument to verify that it's the same.

PR #297, opened 2021-11-03
Update Accessing SharedPreferences

sarahkoop:
### Summary of changes

 - We updated braintree_android to use EncryptedSharedPreferences and provided a `getInstance` and helper `getString` and `putString` methods for accessing shared preferences. This PR updates drop-in to use those methods to interact with shared preferences

 ### Checklist

- @sarahkoop

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
@@ -363,8 +363,8 @@ public class DropInClient {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


sshropshire:
@sarahkoop I can't recall if this feature is mentioned on the docs site. It might be one we can remove temporarily and add back in `4.1` or something once we fully define its use case.

sarahkoop:
> @sarahkoop I can't recall if this feature is mentioned on the docs site. It might be one we can remove temporarily and add back in `4.1` or something once we fully define its use case.

Yeah I checked we do have a [section in the docs for this,](https://developer.paypal.com/braintree/docs/guides/drop-in/setup-and-integration/android/v4#displaying-the-most-recently-added-payment-method) and it is in the latest iOS drop-in, so I think we should probably keep it for now.

PR #298, opened 2021-11-03
Payment Method Fragment Accessibility

sarahkoop:
### Summary of changes

 - Add header to Drop-in with Cancel button (aligns with iOS)
 - Remove cancel button from background view
 - Fetch vaulted payment methods and supported payment methods before displaying drop-in sheet to avoid unexpected UI changes when vaulted payment methods are loaded.
 - Show loading indicator, header, and cancel button while fetching vaulted/supported payment methods

**Before**
Screen Shot 2021-11-03 at 11 43 24 AM
Screen Shot 2021-11-03 at 11 43 36 AM

**After**
Screen Shot 2021-11-03 at 11 42 48 AM
Screen Shot 2021-11-03 at 11 42 14 AM

 ### Checklist

 - ~[ ] Added a changelog entry~ (These accessibility fixes aren't tied to any GH issues - should there be a CHANGELOG for this?)

- @jplukarski 
- @sarahkoop

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
@@ -86,14 +86,7 @@ public class DropInActivity extends AppCompatActivity {
 
         dropInViewModel = new ViewModelProvider(this).get(DropInViewModel.class);
         fragmentContainerView = findViewById(R.id.fragment_container_view);
-
-        dropInClient.getSupportedPaymentMethods(this, (paymentMethods, error) -> {
-            if (paymentMethods != null) {
-                dropInViewModel.setSupportedPaymentMethods(paymentMethods);
-            } else {
-                onError(error);
-            }
-        });
+        fetchPaymentMethods();
 
         getSupportFragmentManager().setFragmentResultListener(DropInEvent.REQUEST_KEY, this,
                 (requestKey, result) -> onDropInEvent(DropInEvent.fromBundle(result)));
@@ -122,7 +115,27 @@ public class DropInActivity extends AppCompatActivity {
 
         showBottomSheet();
     }
+    void fetchPaymentMethods(){
+        dropInClient.getVaultedPaymentMethods(this, (paymentMethodNonceList, error) -> {
+            if (error != null) {
+                onError(error);
+            } else {
+                dropInClient.getSupportedPaymentMethods(this, (paymentMethods, supportedPaymentMethodError) -> {
+                    if (supportedPaymentMethodError != null) {
+                        onError(supportedPaymentMethodError);
+                    }
+                    if (paymentMethods != null) {
+                        dropInViewModel.setSupportedPaymentMethods(paymentMethods);
+                    }
+                    if (paymentMethodNonceList != null){
+                        dropInViewModel.setVaultedPaymentMethods(paymentMethodNonceList);
+                    }
+                    dropInViewModel.setHasFetchedPaymentMethods(true);
+                });
+            }
+        });
 
+    }
     @VisibleForTesting
     void finishDropInWithError(Exception e) {
         setResult(RESULT_FIRST_USER, new Intent().putExtra(DropInResult.EXTRA_ERROR, e));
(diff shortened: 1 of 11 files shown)


Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -122,7 +115,30 @@ public void handleOnBackPressed() {
...
+                    if (paymentMethods != null) {
+                        dropInViewModel.setSupportedPaymentMethods(paymentMethods);
+                    } else {
+                        onError(new BraintreeException("No payment methods supported"));
+                        return;
+                    }
+                    if (paymentMethodNonceList != null){
+                        dropInViewModel.setVaultedPaymentMethods(paymentMethodNonceList);
+                    }
+                    dropInViewModel.setHasFetchedPaymentMethods(true);
```
sarahkoop, line 136: We added `hasFetchedPaymentMethods` to `DropInViewModel` to allow updating the UI when both supported payment methods and vaulted payment methods have finished fetching.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -286,18 +291,6 @@ private void finishDropInWithPendingResult(boolean cancelPendingActivityAnimatio
         }
     }
 
-    private void updateVaultedPaymentMethodNonces(boolean refetch) {
```
sarahkoop, line 289: We removed this because it was only being called immediately before `onError` which dismisses drop-in. Since we are re-fetching the vaulted payment methods in `onCreate`, it doesn't make sense to update these here.

Drop-In/src/main/java/com/braintreepayments/api/SupportedPaymentMethodsFragment.java
```diff
@@ -111,19 +106,22 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
             }
         });
 
+        dropInViewModel.getUserCanceledError().observe(getViewLifecycleOwner(), exception -> {
```
sarahkoop, line 109: We discovered a bug in the 3DS vaulted payment cancel flow - when a user selects a vaulted 3DS card, then cancels from the 3DS flow, the loading indicator is never removed. This update allows the `SupportedPaymentMethodsFragment` to listen for those cancellations and update the UI.

Drop-In/src/main/java/com/braintreepayments/api/SupportedPaymentMethodsFragment.java
```diff
@@ -111,19 +106,22 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+        cancelButton.setOnClickListener(v -> dropInViewModel.setBottomSheetState(BottomSheetState.HIDE_REQUESTED));
         sendAnalyticsEvent("appeared");
         return view;
     }
 
     @Override
     public void onResume() {
         super.onResume();
-
-        if (viewState == ViewState.LOADING && hasSupportedPaymentMethods()) {
```
sarahkoop, line 124: This didn't seem to be necessary - the loading/cancellation flows all seem to be working correctly without this logic. Do we need it?

Drop-In/src/main/java/com/braintreepayments/api/SupportedPaymentMethodsFragment.java
```diff
@@ -201,7 +199,7 @@ private void showVaultedPaymentMethods() {
         }
         
         if (paymentMethodNonces != null && paymentMethodNonces.size() > 0) {
-            supportedPaymentMethodsHeader.setText(R.string.bt_other);
+            supportedPaymentMethodsHeader.setVisibility(View.VISIBLE);
```
sarahkoop, line 202: We updated the supported payment methods header to start with text "Other" and be invisible if no vaulted payment methods exist - this is because there is now a header that says "Select Payment Method" that is always shown, so we only need the "Other" section header when vaulted payment methods exist. See screenshots in PR description.

Drop-In/src/main/res/layout/bt_fragment_supported_payment_methods.xml
```diff
@@ -11,85 +11,116 @@
...
-            android:id="@+id/bt_select_payment_method_loader"
-            android:layout_width="@dimen/bt_progress_bar_diameter"
-            android:layout_height="@dimen/bt_progress_bar_diameter"
-            android:layout_gravity="center"
-            android:indeterminate="true" />
+            android:layout_height="match_parent"
+            android:background="@color/bt_base_background"
+            android:orientation="vertical">
+
+            <ProgressBar
```
sarahkoop, line 116: For some reason - moving this progress bar and it's wrapper within the linear layout that includes the cancel button and select payment method header messes up the centering. Now the progress bar is centered vertically, but not horizontally. Keeping this PR in draft until we can figure out the centering.

sarahkoop:
Closing this for now - putting it off until we can determine approach for nonce caching

PR #299, opened 2021-11-08
Remove Jank

sshropshire:
### Summary of changes

 - Remove initial jank caused by fetching supported payment methods too early
 - Fix logic error causing default animation to be canceled incorrectly

 ### Checklist


Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -261,7 +253,7 @@ private void onDidShowBottomSheet() {
     }
 
     private void onDidHideBottomSheet() {
-        finishDropInWithPendingResult(true);
+        finishDropInWithPendingResult(false);
```
sarahkoop, line 256: I think this needs to be true - when it's false I see that weird UI bug that looks like the toolbar is sliding down with drop-in. If it does get set back to true, it looks like it's true in all usages - do we still need this true/false parameter at all?

sshropshire: Oh right yeah there we may have to add an additional piece of logic to force a fade-out animation. I have an idea of how to make it more expressive.

sshropshire: Ended up adding an enum. The default system animation for an activity exit might cause the toolbar bug to occur. If we explicitly declare a `fade_out` that should resolve the issue.

Drop-In/src/main/java/com/braintreepayments/api/DropInExitTransition.java
```diff
@@ -0,0 +1,6 @@
+package com.braintreepayments.api;
+
+public enum DropInExitTransition {
```
sarahkoop, line 3:
```suggestion
enum DropInExitTransition {
```

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -87,14 +87,6 @@ protected void onCreate(Bundle savedInstanceState) {
         dropInViewModel = new ViewModelProvider(this).get(DropInViewModel.class);
         fragmentContainerView = findViewById(R.id.fragment_container_view);
 
-        dropInClient.getSupportedPaymentMethods(this, (paymentMethods, error) -> {
```
crookedneighbor, line 90: Is this handled elsewhere and it was just duplicated here? or what?

sshropshire: This was an optimistic prefetch. It still happens [here](https://github.com/braintree/braintree-android-drop-in/blob/f9ad3aca587509a1138c73afbdd80d8bd7970dbd/Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java#L255) when the bottom sheet is presented.

We may add a `prefetch()` or `warmup()` method on `DropInClient` to allow for explicit prefetching in the future, but at the moment we've punted on supported / vaulted payment method caching for a post GA release.

Drop-In/src/test/java/com/braintreepayments/api/DropInActivityTest.kt
```diff
@@ -205,6 +204,7 @@ class DropInActivityTest {
     fun onActivityResult_whenResultIsUserCanceledException_setsUserCanceledErrorInViewModel() {
         val error = UserCanceledException("User canceled 3DS.")
         val dropInClient = MockDropInClientBuilder()
+            .authorization(authorization)
```
Epreuve, line 207: What's the story with this little guy?  An artifact from a past test refactor?

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -377,8 +373,8 @@ private void animateBottomSheetClosedAndFinishDropInWithResult(DropInResult drop
             // when the bottom sheet transitions to the "HIDDEN" state; the activity will finish
             dropInViewModel.setBottomSheetState(BottomSheetState.HIDE_REQUESTED);
         } else {
-            // no need to animate bottom sheet hidden; finish activity with default Android animation
-            finishDropInWithPendingResult(false);
+            // no need to animate activity hidden since bottom sheet is not visible
+            finishDropInWithPendingResult(DropInExitTransition.NO_ANIMATION);
```
scannillo, line 377: I'm a little confused why this would ever be a scenario? If we're delivering a DropInResult, wouldn't the sheet always be presented? I'm a little rusty here and would appreciate some additional context I am probably missing!

sshropshire: Yeah that's a good call. `AddCardFragment` has it's own loader when the user submits their card information, so we don't animate when dismissing the activity. I wonder now though if `DropInExitTransition.DEFAULT` would make more sense here since no animation after tokenizing a card might be too abrupt.

scannillo: Ah gotcha - that makes more sense to me!

scannillo (approved):

PR #306, opened 2021-11-11
Refactor DropInPaymentMethodType

sshropshire:
### Summary of changes

 - Rename `DropInPaymentMethodType` to `DropInPaymentMethod`
 - Refactor internals of `DropInPaymentMethod` to make it a pure value type

 ### Checklist

 - [x] Added a changelog entry


Drop-In/src/androidTest/java/com/braintreepayments/api/SupportedPaymentMethodsFragmentUITest.kt
```diff
@@ -25,10 +25,10 @@ import org.mockito.Mockito.mock
...
 
     private val supportedPaymentMethods = listOf(
-            DropInPaymentMethodType.PAYPAL,
-            DropInPaymentMethodType.PAY_WITH_VENMO,
-            DropInPaymentMethodType.UNKNOWN,
-            DropInPaymentMethodType.GOOGLE_PAY
+            DropInPaymentMethod.PAYPAL,
+            DropInPaymentMethod.PAY_WITH_VENMO,
+            DropInPaymentMethod.UNKNOWN,
+            DropInPaymentMethod.GOOGLE_PAY
```
sshropshire, line 31: `PaymentMethod` is a base class in the core SDK. It might make sense to move this enum into core in the next major version.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -327,11 +334,18 @@ void getSupportedPaymentMethods(final FragmentActivity activity, final GetSuppor
     void getSupportedCardTypes(final GetSupportedCardTypesCallback callback) {
         braintreeClient.getConfiguration((configuration, error) -> {
```
sshropshire, line 335: Moving `PaymentMethod` into core would allow us to create a `BraintreeClient#getSupportedCardTypes` method that removes DropIn's responsibility of parsing configuration for payment methods.

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodInspector.java
```diff
@@ -0,0 +1,112 @@
...
+    private static final String PAYMENT_METHOD_HIPERCARD = "Hipercard";
+    private static final String PAYMENT_METHOD_JCB = "JCB";
+    private static final String PAYMENT_METHOD_MAESTRO = "Maestro";
+    private static final String PAYMENT_METHOD_MASTERCARD = "MasterCard";
+    private static final String PAYMENT_METHOD_PAYPAL = "PayPal";
+    private static final String PAYMENT_METHOD_UNION_PAY = "UnionPay";
+    private static final String PAYMENT_METHOD_VENMO = "Venmo";
+    private static final String PAYMENT_METHOD_VISA = "Visa";
+
+    String getPaymentMethodDescription(PaymentMethodNonce paymentMethodNonce) {
```
sshropshire, line 21: We should move `getPaymentMethodDescription()` and `getPaymentMethod()` methods into `PaymentMethodNonce` for the next major version.

sshropshire (commented):
Left notes on next major version.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -327,11 +334,18 @@ void getSupportedPaymentMethods(final FragmentActivity activity, final GetSuppor
...
             if (configuration != null) {
-                List<String> supportedCardTypes = new ArrayList<>(configuration.getSupportedCardTypes());
+                Set<CardType> supportedCardTypes = new HashSet<>();
```
sarahkoop, line 337: Since we are converting supportedCardTypes to an ArrayList to return it, can we just initialize it as an ArrayList here instead?

sshropshire: Yeah good call. I did that to filter out duplicates, but thinking about it `parseCardType` returns `null` for `DropInPaymentMethod.UNKNOWN`. Unless it's duplicated in the gateway that scenario won't happen.

sarahkoop (approved):
A few comments, but looks good when those are resolved.

CHANGELOG.md
```diff
@@ -24,8 +24,7 @@
     * Make `VaultedPaymentMethodSelectedListener` package-private
   * Venmo
     * Remove `vaultVenmo` on `DropInRequest` in favor of setting `shouldVault` on `VenmoRequest`
-  * DropInPaymentMethodType
-    * Rename `DropInPaymentMethodType.GOOGLE_PAYMENT` to `DropInPaymentMethodType.GOOGLE_PAY`
+  * Rename `DropInPaymentMethodType` to `DropInPaymentMethod`
```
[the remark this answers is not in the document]
sshropshire: Yeah I agree this stood out when editing the CHANGELOG. I'll add back a header.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -428,4 +441,9 @@ void getVaultedPaymentMethods(final FragmentActivity activity, final GetPaymentM
             });
         });
     }
+
+    void setLastUsedPaymentMethodType(PaymentMethodNonce paymentMethodNonce) {
+        Context context = braintreeClient.getApplicationContext();
```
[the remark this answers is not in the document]
sshropshire: Yeah I agree. I know we talked about `getApplicationContext` on `BraintreeClient` being less than ideal. I'll refactor to pass in a `Context` since it's internal.

Drop-In/src/main/java/com/braintreepayments/api/DropInPaymentMethod.java
```diff
@@ -0,0 +1,65 @@
+package com.braintreepayments.api;
+
+import com.braintreepayments.api.dropin.R;
+import com.braintreepayments.cardform.utils.CardType;
+
+import androidx.annotation.DrawableRes;
+import androidx.annotation.StringRes;
+
+public enum DropInPaymentMethod {
```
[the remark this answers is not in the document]
sshropshire: 🙏

Drop-In/src/main/java/com/braintreepayments/api/DropInPaymentMethod.java
```diff
@@ -0,0 +1,65 @@
...
+    MAESTRO(CardType.MAESTRO.getFrontResource(), R.drawable.bt_ic_vaulted_maestro, R.string.bt_descriptor_maestro),
+    MASTERCARD(CardType.MASTERCARD.getFrontResource(), R.drawable.bt_ic_vaulted_mastercard, R.string.bt_descriptor_mastercard),
+    PAYPAL(R.drawable.bt_ic_paypal, R.drawable.bt_ic_vaulted_paypal, R.string.bt_descriptor_paypal),
+    VISA(CardType.VISA.getFrontResource(), R.drawable.bt_ic_vaulted_visa, R.string.bt_descriptor_visa),
+    VENMO(R.drawable.bt_ic_venmo, R.drawable.bt_ic_vaulted_venmo, R.string.bt_descriptor_pay_with_venmo),
+    UNIONPAY(CardType.UNIONPAY.getFrontResource(), R.drawable.bt_ic_vaulted_unionpay, R.string.bt_descriptor_unionpay),
+    HIPER(CardType.HIPER.getFrontResource(), R.drawable.bt_ic_vaulted_hiper, R.string.bt_descriptor_hiper),
+    HIPERCARD(CardType.HIPERCARD.getFrontResource(), R.drawable.bt_ic_vaulted_hipercard, R.string.bt_descriptor_hipercard),
+    UNKNOWN(CardType.UNKNOWN.getFrontResource(), R.drawable.bt_ic_vaulted_unknown, R.string.bt_descriptor_unknown),
+    ;
```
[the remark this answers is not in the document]
sshropshire: Yeah good call. I ran the auto-formatter on this it can be better.

Drop-In/src/main/java/com/braintreepayments/api/DropInPaymentMethod.java
```diff
@@ -0,0 +1,65 @@
...
+
+    DropInPaymentMethod(@DrawableRes int drawable, @DrawableRes int vaultedDrawable, @StringRes int localizedName) {
+        this.drawable = drawable;
+        this.vaultedDrawable = vaultedDrawable;
+        this.localizedName = localizedName;
+    }
+
+    /**
+     * @return An id representing a {@link android.graphics.drawable.Drawable} icon for the current
+     * {@link DropInPaymentMethod}.
```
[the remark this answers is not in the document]
sshropshire: Yeah I'll update these to fit the documentation context better.

Drop-In/src/main/java/com/braintreepayments/api/DropInSharedPreferences.java
```diff
@@ -0,0 +1,63 @@
...
+    ) {
+        this.braintreeSharedPreferences = braintreeSharedPreferences;
+        this.paymentMethodInspector = paymentMethodInspector;
+    }
+
+    DropInPaymentMethod getLastUsedPaymentMethod(Context context) {
+        String paymentMethodName = braintreeSharedPreferences.getString(
+                context, LAST_USED_PAYMENT_METHOD, null);
+        if (paymentMethodName != null) {
+            // TODO: catch exception when enum cannot be found with name
```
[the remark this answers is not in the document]
sshropshire: Nope! Good catch. I'll move the comment into the catch block.

Drop-In/src/main/java/com/braintreepayments/api/DropInSharedPreferences.java
```diff
@@ -0,0 +1,63 @@
...
+        return null;
+    }
+
+    void setLastUsedPaymentMethod(Context context, PaymentMethodNonce paymentMethodNonce) {
+        DropInPaymentMethod paymentMethod =
+            paymentMethodInspector.getPaymentMethod(paymentMethodNonce);
+
+        if (paymentMethod != null) {
+            String value = paymentMethod.name();
+            braintreeSharedPreferences.putString(context, LAST_USED_PAYMENT_METHOD, value);
```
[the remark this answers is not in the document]
sshropshire: Yeah the canonical name thing felt like a hack after refactoring. We're restoring an enum here and those names are bound to semantic versioning.

It also allowed us to remove the concept of canonical names. They were primarily used to map gateway `paymentMethodType` strings to the enums. Another one of the reasons I feel like `PaymentMethodType` should be a core SDK concept.

Drop-In/src/test/java/com/braintreepayments/api/PaymentMethodInspectorTest.java
```diff
@@ -0,0 +1,228 @@
...
+    @Before
+    public void beforeEach() {
+        cardNonce = mock(CardNonce.class);
+        payPalNonce = mock(PayPalAccountNonce.class);
+        venmoNonce = mock(VenmoAccountNonce.class);
+        googlePayNonce = mock(GooglePayCardNonce.class);
+
+        sut = new PaymentMethodInspector();
+    }
+
```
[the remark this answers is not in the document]
sshropshire: Good catch. They are all in there except they're named `getCardType` I'll update.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
Epreuve: Is there any equivalent currently in core for either Android or iOS?  JS does have one in core, and I agree it would make more sense to have it there, so it's both accessible to use in Drop-in, but can serve to allow merchants more control over more custom integrations.

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodInspector.java
Epreuve: Is this approach then to avoid further delay in GA by not relying on an upstream change/merge?  Just trying to weigh the pros/cons on this one.

Epreuve (approved):
Just a couple clarifying questions, but overall looks like a solid clean up.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
scannillo: Look at this on iOS, turns out we currently aren't exposing `supportedCardTypes` at all. ([This](https://github.com/braintree/braintree-ios-drop-in/blob/febbaa5afb38853ff9a5bf35b0a444518777a01e/Sources/BraintreeDropIn/Helpers/BTConfiguration%2BDropIn.h) is not public)

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -327,9 +333,16 @@ void getSupportedPaymentMethods(final FragmentActivity activity, final GetSuppor
...
+                for (String cardTypeAsString : configuration.getSupportedCardTypes()) {
+                    CardType cardType = paymentMethodInspector.parseCardType(cardTypeAsString);
+                    if (cardType != null) {
+                        supportedCardTypes.add(cardType);
+                    }
+                }
+
                 if (!configuration.isUnionPayEnabled()) {
-                    supportedCardTypes.remove(DropInPaymentMethodType.UNIONPAY.getCanonicalName());
+                    supportedCardTypes.remove(CardType.UNIONPAY);
```
scannillo, line 345: If union pay is not enabled, and we try to remove this value from the list, but it doesn't exist in the list - wouldn't we get a crash here?

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
sshropshire: > Is there any equivalent currently in core for either Android or iOS?

@Epreuve we don't at the moment. A lot of the BT configuration accessors are plain `String` types. In this case, it feels like additional parsing logic in core could lessen the burden on all core-consuming modules.

> Agreed we should align on this capability across platforms.

@scannillo yeah alignment would be awesome. We will have to do a major version bump in the near future because of Cardinal. Should be a good opportunity to align the two.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
sshropshire: Java `List` will actually no-op in this case. It would throw though if we were removing by index and the index was out of bounds.

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodInspector.java
sshropshire: Yeah pretty much. It is internal though, so if we decided to add those methods in core we can gradually remove this class without breaking changes.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
scannillo: Awesome - thanks for confirming

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -282,43 +289,42 @@ void getSupportedPaymentMethods(final FragmentActivity activity, final GetSuppor
...
             Set<String> supportedCardTypes =
                     new HashSet<>(configuration.getSupportedCardTypes());
             if (!configuration.isUnionPayEnabled()) {
-                supportedCardTypes.remove(DropInPaymentMethodType.UNIONPAY.getCanonicalName());
+                supportedCardTypes.remove(CARD_TYPE_UNION_PAY);
             }
-            if (supportedCardTypes.size() > 0) {
-                availablePaymentMethods.add(DropInPaymentMethodType.UNKNOWN);
+            if (supportedCardTypes.size() > 0) { availablePaymentMethods.add(DropInPaymentMethod.UNKNOWN);
             }
```
scannillo, line 322:
```suggestion
            if (supportedCardTypes.size() > 0) { 
                availablePaymentMethods.add(DropInPaymentMethod.UNKNOWN);
            }
```

Nit

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodInspector.java
```diff
@@ -0,0 +1,112 @@
...
+        } else if (paymentMethodNonce instanceof VenmoAccountNonce) {
+            return PAYMENT_METHOD_VENMO;
+        } else if (paymentMethodNonce instanceof GooglePayCardNonce) {
+            return PAYMENT_METHOD_GOOGLE_PAY;
+        } else {
+            return null;
+        }
+    }
+
+    CardType parseCardType(String cardType) {
```
scannillo, line 86:
```suggestion
    CardType getCardType(String cardType) {
```

Another nitpick - but this naming seems a little more straightforward to me, since we aren't really "parsing" anything

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodInspector.java
```diff
@@ -0,0 +1,112 @@
+package com.braintreepayments.api;
+
+import com.braintreepayments.cardform.utils.CardType;
+
+class PaymentMethodInspector {
```
scannillo, line 5: Curious the reasoning here for making this it's own file vs including within `DropInPaymentMethod`?

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodInspector.java
sshropshire: Yeah def. I'll rename.

Drop-In/src/main/java/com/braintreepayments/api/DropInResult.java
```diff
@@ -27,46 +26,46 @@
...
 
-    void setPaymentMethodNonce(@Nullable PaymentMethodNonce paymentMethodNonce) {
-        setPaymentMethodNonce(paymentMethodNonce, new PaymentMethodNonceInspector());
+    DropInResult setPaymentMethodNonce(@Nullable PaymentMethodNonce paymentMethodNonce) {
+        return setPaymentMethodNonce(paymentMethodNonce, new PaymentMethodInspector());
     }
 
     @VisibleForTesting
-    void setPaymentMethodNonce(@Nullable PaymentMethodNonce paymentMethodNonce, PaymentMethodNonceInspector nonceInspector) {
+    DropInResult setPaymentMethodNonce(@Nullable PaymentMethodNonce paymentMethodNonce, PaymentMethodInspector nonceInspector) {
```
sarahkoop, line 42: I think there were some rebase issues with this class - we removed the builder pattern here in #307

Drop-In/src/main/java/com/braintreepayments/api/DropInResult.java
```diff
@@ -88,21 +87,11 @@ public String getDeviceData() {
         return deviceData;
     }
 
-    /**
-     * @return A {@link String} description of the payment method.
```
sarahkoop, line 92: I think this was also lost in the rebase - we should have this doc string

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -337,6 +351,7 @@ void getSupportedCardTypes(final GetSupportedCardTypesCallback callback) {
         });
     }
 
+
```
sarahkoop, line 354:
```suggestion
```

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodInspector.java
sshropshire: Yeah this sort of became a junk drawer for code that may be more suitable for the core sdk. Another goal was to make `DropInPaymentMethod` more of a pure value type, although it may make sense in `v5` to create a `PaymentMethod` enum in core instead.

Drop-In/src/main/java/com/braintreepayments/api/DropInResult.java
sshropshire: Yeah good call. The rebase had some conflicts I'll update.

Drop-In/src/main/java/com/braintreepayments/api/DropInResult.java
```diff
@@ -55,18 +51,18 @@ void setDeviceData(@Nullable String deviceData) {
...
-    void setPaymentMethodType(@Nullable DropInPaymentMethodType paymentMethodType) {
+    void setPaymentMethodType(DropInPaymentMethod paymentMethodType) {
         this.paymentMethodType = paymentMethodType;
     }
 
     /**
-     * @return The previously used {@link DropInPaymentMethodType} or {@code null} if there was no
-     * previous payment method. If the type is {@link DropInPaymentMethodType#GOOGLE_PAY} the Google
+     * @return The previously used {@link DropInPaymentMethod} or {@code null} if there was no
+     * previous payment method. If the type is {@link DropInPaymentMethod#GOOGLE_PAY} the Android
```
sarahkoop, line 60:
```suggestion
     * previous payment method. If the type is {@link DropInPaymentMethod#GOOGLE_PAY} the Google
```

noticed one other change

PR #307, opened 2021-11-11
General Cleanup

sarahkoop:
### Summary of changes

 - Reviewed public-facing code and made a few classes/methods package-private
 - Reviewed for nullability annotations and doc strings on public methods
 - Removed builder pattern from `DropInResult` - the setters on this class aren't public, but since we removed the builder pattern everywhere else it felt like this should be cleaned up too

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop


sshropshire (approved):
🌊

Epreuve (approved):
🌵

PR #308, opened 2021-11-12
Update Localizations

sarahkoop:
### Summary of changes

 - Update localizations from translations repository
 - Add four new strings for accessibility screen reading
 - Add support for Czech, Greek, Finnish, Hebrew, Hungrian, and Slovak localization
 - Remove support for Turkish localization

**Note**: After running the script to pull the current strings from the translations repo, I noticed that we had manually added Hiper and Hipercard, and remove Android Pay directly in this repository. I re-made those changes here, but we will need to open a new PR in the translations repo to get the strings in sync.

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop


Drop-In/src/main/res/values-en/strings.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sarahkoop, line 1: I'm not sure if we need this file - it's English, but slightly different than the source English file. The translations repo had both the source file (which correlates to values/strings.xml) and this values-en file. Should we have both?

sshropshire: It should be safe to delete. Android will default to `Drop-In/src/main/res/values/strings.xml` if a locale translation isn't specified.

sshropshire (approved):

sshropshire:
> **Note**: After running the script to pull the current strings from the translations repo, I noticed that we had manually added Hiper and Hipercard, and remove Android Pay directly in this repository. I re-made those changes here, but we will need to open a new PR in the translations repo to get the strings in sync.

This was probably me haha. I didn't know there was a script to update translations 😬

Drop-In/src/main/res/values-en/strings.xml
sarahkoop: 👍 Removed

PR #315, opened 2021-11-30
Support Activity Result API

### LINKED ISSUE #313: dropInClient.launchDropInForResult allow Fragments or better registerForActivityResult
### General information

* SDK/Library version: 6.0.0.
* Environment: sandbox
* Android Version and Device: any

### Issue description
`dropInClient.launchDropInForResult()` requires Activity. But in this case `startActivityForResult` is called on activity, not in attached fragment. So result is returned to activity and not into active fragment.

### Expected behavior
`launchDropInForResult` accept Fragments, so we can expect result in Fragment and not only in activity.

Or even better accept Activity Result APIs lambda function to handle result in callback through `registerForActivityResult`. More info about Activity Result APIs:
https://developer.android.com/training/basics/intents/result
https://medium.com/e-legion/the-right-way-to-get-a-result-part-i-activity-result-api-6efbcaa5600d

sarahkoop:
### Summary of changes

 - Add support for Activity Result APIs and launching Drop-in via Fragment (fixes #313)

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
@@ -3,6 +3,7 @@ package com.braintreepayments.demo;
 import static android.view.View.GONE;
 import static android.view.View.VISIBLE;
 
+import android.app.Activity;
 import android.content.Intent;
 import android.os.Bundle;
 import android.view.View;
@@ -10,10 +11,15 @@ import android.widget.Button;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import androidx.activity.result.ActivityResult;
+import androidx.activity.result.ActivityResultCallback;
+import androidx.activity.result.ActivityResultLauncher;
+import androidx.activity.result.contract.ActivityResultContracts;
 import androidx.annotation.NonNull;
 import androidx.cardview.widget.CardView;
 
 import com.braintreepayments.api.CardNonce;
+import com.braintreepayments.api.DropInActivityContract;
 import com.braintreepayments.api.DropInClient;
 import com.braintreepayments.api.DropInPaymentMethod;
 import com.braintreepayments.api.DropInRequest;
@@ -53,6 +59,7 @@ public class MainActivity extends BaseActivity {
     private Button purchaseButton;
 
     private boolean purchased = false;
+    private ActivityResultLauncher<DropInClient> resultLauncher;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -75,6 +82,19 @@ public class MainActivity extends BaseActivity {
                 nonce = savedInstanceState.getParcelable(KEY_NONCE);
             }
         }
+
+        resultLauncher = registerForActivityResult(new DropInActivityContract(), result -> {
+            if (result != null) {
+                if (result.getDropInResult() != null) {
+                    displayResult(result.getDropInResult());
+                } else if (result.getError() != null) {
+                    showDialog(result.getError().getMessage());
+                }
+            } else {
+                // User cancelled drop in
+            }
+
+        });
     }
 
     @Override
@@ -96,7 +116,7 @@ public class MainActivity extends BaseActivity {
     }
 
     public void launchDropIn(View v) {
-        dropInClient.launchDropInForResult(this, DROP_IN_REQUEST);
+        resultLauncher.launch(dropInClient);
     }
 
     private ThreeDSecureRequest demoThreeDSecureRequest() {
@@ -148,19 +168,19 @@ public class MainActivity extends BaseActivity {
         }
     }
 
-    @Override
-    public void onActivityResult(int requestCode, int resultCode, Intent data) {
-        super.onActivityResult(requestCode, resultCode, data);
-
-        if (resultCode == RESULT_OK) {
-            DropInResult result = data.getParcelableExtra(DropInResult.EXTRA_DROP_IN_RESULT);
-            displayResult(result);
-            purchaseButton.setEnabled(true);
-        } else if (resultCode != RESULT_CANCELED) {
-            showDialog(((Exception) data.getSerializableExtra(DropInResult.EXTRA_ERROR))
-                    .getMessage());
-        }
-    }
+//    @Override
+//    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+//        super.onActivityResult(requestCode, resultCode, data);
+//
+//        if (resultCode == RESULT_OK) {
+//            DropInResult result = data.getParcelableExtra(DropInResult.EXTRA_DROP_IN_RESULT);
+//            displayResult(result);
+//            purchaseButton.setEnabled(true);
+//        } else if (resultCode != RESULT_CANCELED) {
+//            showDialog(((Exception) data.getSerializableExtra(DropInResult.EXTRA_ERROR))
+//                    .getMessage());
+//        }
+//    }
 
     @Override
     protected void reset() {
--- a/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
@@ -3,7 +3,6 @@ package com.braintreepayments.demo;
 import static android.view.View.GONE;
 import static android.view.View.VISIBLE;
 
-import android.app.Activity;
 import android.content.Intent;
 import android.os.Bundle;
 import android.view.View;
@@ -11,10 +10,7 @@ import android.widget.Button;
 import android.widget.ImageView;
 import android.widget.TextView;
 
-import androidx.activity.result.ActivityResult;
-import androidx.activity.result.ActivityResultCallback;
 import androidx.activity.result.ActivityResultLauncher;
-import androidx.activity.result.contract.ActivityResultContracts;
 import androidx.annotation.NonNull;
 import androidx.cardview.widget.CardView;
 
@@ -93,7 +89,6 @@ public class MainActivity extends BaseActivity {
             } else {
                 // User cancelled drop in
             }
-
         });
     }
 
@@ -167,20 +162,6 @@ public class MainActivity extends BaseActivity {
             purchaseButton.setEnabled(true);
         }
     }
-
-//    @Override
-//    public void onActivityResult(int requestCode, int resultCode, Intent data) {
-//        super.onActivityResult(requestCode, resultCode, data);
-//
-//        if (resultCode == RESULT_OK) {
-//            DropInResult result = data.getParcelableExtra(DropInResult.EXTRA_DROP_IN_RESULT);
-//            displayResult(result);
-//            purchaseButton.setEnabled(true);
-//        } else if (resultCode != RESULT_CANCELED) {
-//            showDialog(((Exception) data.getSerializableExtra(DropInResult.EXTRA_ERROR))
-//                    .getMessage());
-//        }
-//    }
 
     @Override
     protected void reset() {
--- a/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
@@ -10,13 +10,12 @@ import android.widget.Button;
 import android.widget.ImageView;
 import android.widget.TextView;
 
-import androidx.activity.result.ActivityResultLauncher;
 import androidx.annotation.NonNull;
 import androidx.cardview.widget.CardView;
 
 import com.braintreepayments.api.CardNonce;
-import com.braintreepayments.api.DropInActivityContract;
 import com.braintreepayments.api.DropInClient;
+import com.braintreepayments.api.DropInLauncher;
 import com.braintreepayments.api.DropInPaymentMethod;
 import com.braintreepayments.api.DropInRequest;
 import com.braintreepayments.api.DropInResult;
@@ -29,6 +28,7 @@ import com.braintreepayments.api.PostalAddress;
 import com.braintreepayments.api.ThreeDSecureAdditionalInformation;
 import com.braintreepayments.api.ThreeDSecurePostalAddress;
 import com.braintreepayments.api.ThreeDSecureRequest;
+import com.braintreepayments.api.UserCanceledException;
 import com.braintreepayments.api.VenmoAccountNonce;
 import com.braintreepayments.api.VenmoPaymentMethodUsage;
 import com.braintreepayments.api.VenmoRequest;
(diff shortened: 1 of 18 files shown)


Drop-In/src/main/java/com/braintreepayments/api/DropInActivityContract.java
```diff
@@ -0,0 +1,41 @@
...
+                .putExtra(EXTRA_AUTHORIZATION, dropInClient.getAuthorization().toString());
+        return intent;
+    }
+
+    @Override
+    public DropInResult2 parseResult(int resultCode, @Nullable Intent intent) {
+        if (resultCode == RESULT_OK || resultCode == RESULT_FIRST_USER) {
+            return intent.getParcelableExtra(DropInResult2.EXTRA_DROP_IN_RESULT_2);
+        }
+        return null;
```
sarahkoop, line 39: This does nothing for cancelled results, which is currently how we handle it - but not sure if the docs story for that is ok (see README)

Drop-In/src/main/java/com/braintreepayments/api/DropInResult2.java
```diff
@@ -0,0 +1,61 @@
+package com.braintreepayments.api;
+
+import android.os.Parcel;
+import android.os.Parcelable;
+
+import androidx.annotation.Nullable;
+
+public class DropInResult2 implements Parcelable  {
```
sarahkoop, line 8: Looking for input here. The activity contract requires a single result object. We could add error as a field on the existing `DropInResult` object, but that could cause confusion for existing integrations not using this Activity Result API. I decided to create a wrapper result that contains a `DropInResult` or `Exception` to keep it simple and separate the integrations. Couldn't think of a good name for this new result so looking for suggestions on that as well.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -115,7 +115,13 @@ public void handleOnBackPressed() {
 
     @VisibleForTesting
     void finishDropInWithError(Exception e) {
-        setResult(RESULT_FIRST_USER, new Intent().putExtra(DropInResult.EXTRA_ERROR, e));
+        Intent intent = new Intent();
+        intent.putExtra(DropInResult.EXTRA_ERROR, e);
+
+        DropInResult2 result = new DropInResult2();
+        result.setError(e);
+        intent.putExtra(DropInResult2.EXTRA_DROP_IN_RESULT_2, result);
```
sarahkoop, line 123: I decided to duplicate the logic on setting the result as an extra on the intent. This ensures that existing integrations can access the drop in result or error using the same extra keys as before, and in `DropInActivityContract` we can use the new key to access the `DropInResult2` to be returned to the merchants using the new integration.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivityContract.java
sshropshire: Yeah that makes sense to me.

Drop-In/src/main/java/com/braintreepayments/api/DropInResult2.java
sshropshire: I think this is a good call. `DropInActivityResult` might be an option as well, though it's really similar to `DropInResult`. If this is the recommended route though maybe we can mark the previous result class a deprecated? Not sure how that would work or if it makes sense but we have a few options.

sshropshire:
@sarahkoop This is a good call. I like where it's at fundamentally. There is also [this](https://developer.android.com/training/basics/intents/result?hl=fr#separate) there could be an opportunity to streamline the API a bit and make it more BT style than Jetpack style.

sarahkoop:
> @sarahkoop This is a good call. I like where it's at fundamentally. There is also [this](https://developer.android.com/training/basics/intents/result?hl=fr#separate) there could be an opportunity to streamline the API a bit and make it more BT style than Jetpack style.

@sshropshire Merchants should be able to choose if they want to use to use the `registerForActivityResult` method from within a Fragment or Activity, or create a custom lifecycle observer and register the `DropInActivityContract` within the observer instead. The docs story for `registerForActivityResult` seemed a little clearer to me so I used that in the README - but do you think we should document how the lifecycle observer pattern would work instead/in addition?

Drop-In/src/main/java/com/braintreepayments/api/DropInResult2.java
sarahkoop: Hmm.. I marked it as deprecated, but am also wondering if that will cause confusion for merchants unless we update our docs and migration guide to essentially ignore the `DropInClient#launchDropInForResult` option. It might be a big lift to ask merchants to upgrade to v6 and adopt the new Activity Result APIs. Maybe we should leave them both as options for now and not deprecate?

sshropshire: Ah yeah that's a good point. Maybe we could keep `DropInResult` then and add the `error` field? We can still keep the error `Intent` extra while also adding it to the `DropInResult` parcelable. A bit redundant, but it would prevent a breaking change.

sshropshire:
> do you think we should document how the lifecycle observer pattern would work instead/in addition?

If possible it'd be dope if we owned the callback signature so that the merchant is unaware of `registerActivityForResult`. It is a good callout that the merchant has a choice between `Fragment` and `FragmentActivity`.

I'm wondering if there's a way we could have a `DropInClient#registerActivityForResult` method with `Fragment` and `FragmentActivity` overloads that calls the androidx `registerActivityForResult()` under the hood? The name could even be something more in our domain e.g. `registerForDropInResult`.

Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
```diff
@@ -75,6 +78,18 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+        resultLauncher = registerForActivityResult(new DropInActivityContract(), result -> {
+            if (result != null) {
+                if (result.getDropInResult() != null) {
+                    displayResult(result.getDropInResult());
+                } else if (result.getError() != null) {
+                    showDialog(result.getError().getMessage());
+                }
+            } else {
+                // User cancelled drop in
```
scannillo, line 90: Thank you for this comment 👏

Drop-In/src/main/java/com/braintreepayments/api/DropInActivityContract.java
```diff
@@ -0,0 +1,45 @@
...
+import android.content.Intent;
+import android.os.Bundle;
+
+import androidx.activity.result.contract.ActivityResultContract;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+
+/**
+ * A contract used to launch a {@link DropInActivity} with a {@link DropInClient} input and {@link DropInResult2} output.
```
scannillo, line 21: Is `contract` common terminology for this sort of flow on Android?

Drop-In/src/main/java/com/braintreepayments/api/DropInResult2.java
scannillo: I also like the idea of creating this wrapper class and not muddying up the existing `DropInResult`. 

Another option is that we mark every component of the old flow (non activity result API) as deprecated, and then we do include these updates into the existing `DropInResult`. 

Also I like Steven's naming suggestion of `DropInActivityResult` if we are going forward with a separate class. 

Either way, I think we should definitely deprecate the old flow and update our docs at the same time. I think having 2 flows that do the same thing is confusing. And deprecating doesn't cause anything to break, just signals to merchants "hey you should update this sooner than later".

README.md
```diff
@@ -148,6 +148,47 @@ dropInClient.fetchMostRecentPaymentMethod(this) { dropInResult, error ->
...
+Within your Activity or Fragment where you will launch Drop-in, create an `ActivityResultLauncher`:
+
+```kotlin
+class MyFragment : Fragment() {
+    
+    private lateinit var resultLauncher: ActivityResultLauncher<DropInClient>
+    
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        resultLauncher = registerForActivityResult(DropInActivityContract()) { result ->
```
scannillo, line 171: 👏

Drop-In/src/main/java/com/braintreepayments/api/DropInResult2.java
sarahkoop: Ok I am going to go with the wrapper class and rename it to `DropInActivityResult`, and deprecate `DropInClient#launchDropInForResult` 👍

Drop-In/src/main/java/com/braintreepayments/api/DropInActivityContract.java
sarahkoop: I haven't seen it before, but I based this doc string on the class it extends (`ActivityResultContract`), which uses the `contract` terminology.

Drop-In/src/main/java/com/braintreepayments/api/DropInResult2.java
scannillo: Shouldn't we also deprecate the entire `DropInResult` too?

sarahkoop: I don't think it should be deprecated since `DropInActivityResult` contains a `DropInResult` or an `Exception` - that `DropInResult` still holds the actual payment method type, nonce, etc. In the deprecated flow the `DropInResult` would be accessed from an intent extra in `onActivityResult`, and now it will be accessed from the `DropInActivityResult` produced by the `DropInActivityContract`. Do you think that's clear enough?

scannillo: Let me know if that idea makes sense - but basically it's in favor of duplicating now, so that the deprecation / update path is more clear.

sarahkoop: Hmm I still don't think I understand the need to deprecate `DropInResult` at all, it's not really impacted by this change. The only reason this `DropInActivityResult` class is added is because this approach requires a single return object, and it seemed cleaner for the merchant code to have the single object either have a non-null result, or a non-null error. We could alternatively just add the error as a parameter on the original `DropInResult`, but then the null-checking in the merchant code isn't quite as simple. It's not a huge difference but I think it would be something like this:

```
        resultLauncher = registerForActivityResult(DropInActivityContract()) { result ->
            result?.let {
                it.dropInResult?.let { dropInResult ->
                    // handle result accessing dropInResult.paymentMethodNonce?, dropInResult.paymentMethodType?, etc
                }
                it.error?.let { dropInError ->
                    // handle error
                }
            }
            // user cancelled drop in
        }
```
vs: 
```
        resultLauncher = registerForActivityResult(DropInActivityContract()) { result ->
            result?.let {
                it.paymentMethodNonce?.let { paymentMethodNonce ->
                    // handle result using fields on outer-scoped it (it.paymentMethodType?, it.paymentMethodDescription?) and paymentMethodNonce
                }
                it.error?.let { dropInError ->
                    // handle error
                }
            }
            // user cancelled drop in
        }
```

Maybe the difference isn't a big enough reason to add this wrapper type. What do you think?

sarahkoop: @sshropshire Sammy and I discussed this a bit more and decided that the single `DropInActivityResult` object makes sense, and we will keep the `DropInResult` not deprecated since it contains the result information. We are now considering the cancelation handling with this new flow. Previously, in `onActivityResult`, the merchant would received an explicit `RESULT_CANCELED` result code so that they could handle the cancellation case. In this flow, we are kind of lumping cancellations into any null result, and aren't explicitly telling a merchant that the result was cancelled.  There are a few ways we could handle this:
1) Keep it as is, and allow merchants to handle cancellation along with any other null result outcomes (but give no explicit cancellation information)
2) Add an `isCanceled` boolean to the `DropInActivityResult` 
3) Return cancelation as a `UserCanceledException` in the `error` field already on `DropInActivityResult`

Do you have an opinion on which approach would be best?

sshropshire: I think any option could work. If possible I'd wrap the androidx `registerForActivityResult` method to give us ownership of the callback method signature. Then we could return`(DropInResult, Error)` to the merchant without them knowing we're using the register activity API under the hood.

sarahkoop: Since `registerForActivityResult` is a method on `ComponentActivity` and `Fragment` I'm not sure that we could do that. We couldn't allow them to use DropInClient to call it at least. We'd have to create our own Fragment and Activity that merchants would override I think.

sshropshire: Yeah that's true. We would have to have multiple overloads for the start method. Like `launchDropInForResult(ComponentActivity)` and `launchDropInForResult(Fragment)`.

sarahkoop: Ah good idea! I spec'd out what that would look like with a `DropInLauncher` class. It's a single commit (without updating tests) so can revert if we prefer giving merchants full control over the Activity Result API.
Pros to `DropInLauncher`:
- We can keep the `DropInActivityResult` class internal and pass back a tuple (result, error) to a callback similar to all other APIs
- We can make `DropInActivityContract` non-public and hide the actual Activity Result API logic from the merchant

Cons to `DropInLauncher`:
- Merchants need to create both a `DropInClient` and a `DropInLauncher` to launch Drop-in (`DropInClient` can't easily be remove/instantiated within `DropInLauncher` because we need to call `DropInLauncher#registerForActivityResult` before the fragment/activity is created, but we need to give merchants a chance to fetch authorization before creating a `DropInClient`
- We take responsibility for documenting when to call the lifecycle methods (i.e. call `DropInLauncher#registerForActivityResult` in `onCreate` and make sure to call it before you call `DropInLauncher.launchWithDropInClient`) instead of letting Google Activity Result API document the requirements. 

Let me know what you think/if I'm missing anything else with this approach.

scannillo: This is super helpful, thanks for outlining it so clearly. I guess we need to decide which is more important - more granular/specific return type for merchants or more control over activity launch lifecycle.

sshropshire (approved):
🔥

Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
```diff
@@ -75,6 +76,19 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            if (result != null) {
+                displayResult(result);
+            } else if (error != null) {
+                if (error instanceof UserCanceledException) {
+                    // User canceled
+                } else {
+                    showDialog(error.getMessage());
+                }
+            }
+        });
```
sshropshire, line 91: 🚢 &nbsp;👏

sarahkoop:
Closing in favor of #337

PR #320, opened 2021-12-08
Fix Rotation Null Pointer Exception

### LINKED ISSUE #319: java.lang.NullPointerException:BottomSheetPresenter.unbind()
This is how to test the demo application:

image
image

The phone turns on automatic rotation, enters the credit or debit card input page, crashes after the horizontal screen, and exits.
It seems to appear every time.

Logcat Output:
image
 
This is a video recorded:

https://user-images.githubusercontent.com/86334406/145173579-daaa7af4-9f2c-48bb-ba99-6dabc39f8141.mp4

sshropshire:
### Summary of changes

 - Check if `BottomSheetPresenter` is null before unbinding (fixes #319)
 - Set `saveEnabled` to `false` for ViewPager2 until we have explicit support for saving instance state

 ### Checklist

 - [ ] Added a changelog entry

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/BottomSheetFragment.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/BottomSheetFragment.java
@@ -51,7 +51,14 @@ public class BottomSheetFragment extends Fragment implements BottomSheetPresente
 
         View view = inflater.inflate(R.layout.bt_fragment_bottom_sheet, container, false);
         backgroundView = view.findViewById(R.id.background);
+
         viewPager = view.findViewById(R.id.view_pager);
+        // TODO: investigate view pager saveInstanceState restoration
+        viewPager.setSaveEnabled(false);
+
+        // it's best to call bind here before any live data / fragment result observers are registered
+        bottomSheetPresenter = new BottomSheetPresenter();
+        bottomSheetPresenter.bind(this);
 
         FragmentManager childFragmentManager = getChildFragmentManager();
         childFragmentManager.setFragmentResultListener(DropInEvent.REQUEST_KEY, this,
@@ -98,9 +105,6 @@ public class BottomSheetFragment extends Fragment implements BottomSheetPresente
         Button backButton = view.findViewById(R.id.back_button);
         backButton.setOnClickListener(v -> slideDownBottomSheet());
 
-        bottomSheetPresenter = new BottomSheetPresenter();
-        bottomSheetPresenter.bind(this);
-
         return view;
     }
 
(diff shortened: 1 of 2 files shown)


Drop-In/src/main/java/com/braintreepayments/api/BottomSheetFragment.java
```diff
@@ -51,7 +51,14 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
 
         View view = inflater.inflate(R.layout.bt_fragment_bottom_sheet, container, false);
         backgroundView = view.findViewById(R.id.background);
+
         viewPager = view.findViewById(R.id.view_pager);
+        // TODO: investigate view pager saveInstanceState restoration
+        viewPager.setSaveEnabled(false);
```
sarahkoop, line 57: It might make sense to tackle this TODO in this PR too if possible. I think we need to save instance state of the viewPager or if you rotate the screen while in the vault manager screen, it jumps back to the payment method page instead.

sarahkoop: Ok yeah that makes sense

sarahkoop (approved):

Drop-In/src/main/java/com/braintreepayments/api/BottomSheetFragment.java
```diff
@@ -51,7 +51,14 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+
+        // it's best to call bind here before any live data / fragment result observers are registered
```
jplukarski, line 59: Just curious, do you have a TLDR on why it's better to call bind here?

jplukarski (approved):
Just one quick question, otherwise LGTM

braintreeps (approved):
Seems pretty straightforward.  Is there anything setup to track this TODO?

Drop-In/src/main/java/com/braintreepayments/api/BottomSheetFragment.java
sshropshire: Yeah we observe the `ViewModel` for changes in `onCreate`, and some of the `LiveData` in the view model have initial values. When there is initial data, the observer callbacks fire as soon as we call `observe()`, so we need to instantiate `bottomSheetPresenter` early to make sure we don't get null pointer exceptions that cause the app to crash.

sshropshire: The true TLDR is "because Java" lol.

sshropshire:
> Seems pretty straightforward. Is there anything setup to track this TODO?

That's a good call. I'll write something up and reference this PR.

PR #337, opened 2022-03-25
Client Token Provider

sshropshire:
### Summary of changes

 - Add `ClientTokenProvider` integration method to DropIn

 ### Checklist

 - [x] Added a changelog entry


Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -329,15 +334,20 @@ private void showBottomSheet() {
...
-                            Authorization.isTokenizationKey(dropInClient.getAuthorization().toString());
-
-                    CardDetailsFragment cardDetailsFragment = CardDetailsFragment.from(
-                            dropInRequest, cardNumber, configuration, hasTokenizationKeyAuth);
-                    replaceExistingFragment(cardDetailsFragment, CARD_DETAILS_TAG);
+            dropInClient.getAuthorization(new AuthorizationCallback() {
+                @Override
+                public void onAuthorizationResult(@Nullable Authorization authorization, @Nullable Exception authorizationError) {
+                    dropInClient.getConfiguration((configuration, configurationError) -> {
+                        // TODO: determine UX for scenario where authorization or configuration could not be fetched
```
sarahkoop, line 341: We could probably do something like what we did before for `InvalidAuthorization`. Maybe error message like "There was a problem fetching authorization, or the authorization was invalid"?

Drop-In/src/main/java/com/braintreepayments/api/DropInIntentData.java
```diff
@@ -0,0 +1,26 @@
+package com.braintreepayments.api;
+
+public class DropInIntentData {
```
sarahkoop, line 3: I think this can all be package private

Drop-In/src/main/java/com/braintreepayments/api/DropInLifecycleObserver.java
```diff
@@ -0,0 +1,46 @@
...
+import androidx.activity.result.ActivityResultLauncher;
+import androidx.activity.result.ActivityResultRegistry;
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
+import androidx.lifecycle.Lifecycle;
+import androidx.lifecycle.LifecycleEventObserver;
+import androidx.lifecycle.LifecycleOwner;
+
+// NEXT_MAJOR_VERSION: Update to implement DefaultLifeCycleObserver when Java 7 support is explicitly dropped.
+public class DropInLifecycleObserver implements LifecycleEventObserver {
```
sarahkoop, line 13: We can go ahead and just use `DefaultLifeCycleObserver` in Drop-in since we dropped Java 7 with the major version

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -44,10 +49,23 @@
...
+    private DropInListener listener;
+
+    @VisibleForTesting
+    DropInLifecycleObserver observer;
+
+    private static DropInClientParams createDefaultParams(Context context, String authorization, ClientTokenProvider clientTokenProvider, DropInRequest dropInRequest, String sessionId, FragmentActivity activity, Lifecycle lifecycle) {
+
+        BraintreeClient braintreeClient;
+        if (clientTokenProvider != null) {
+            braintreeClient = new BraintreeClient(context, clientTokenProvider);
```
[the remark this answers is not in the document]
sshropshire: Yeah I forgot to add I'll make a PR to add one to core.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Good call! I can finish the activity with the authorization error.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -287,15 +286,21 @@ private void finishDropInWithPendingResult(DropInExitTransition transition) {
...
     private void updateVaultedPaymentMethodNonces(boolean refetch) {
-        if (clientTokenPresent) {
-            dropInClient.getVaultedPaymentMethods(this, (vaultedPaymentMethods, error) -> {
-                if (vaultedPaymentMethods != null) {
-                    dropInViewModel.setVaultedPaymentMethods(vaultedPaymentMethods);
-                } else if (error != null) {
-                    onError(error);
+        dropInClient.getAuthorization(new AuthorizationCallback() {
+            @Override
+            public void onAuthorizationResult(@Nullable Authorization authorization, @Nullable Exception authorizationError) {
```
sarahkoop, line 291: We should handle this authorizationError too

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
```diff
@@ -329,15 +334,21 @@ private void showBottomSheet() {
...
+                    dropInClient.getConfiguration((configuration, configurationError) -> {
+                        if (configuration != null) {
+                            boolean hasTokenizationKeyAuth =
+                                    Authorization.isTokenizationKey(authorization.toString());
+
+                            CardDetailsFragment cardDetailsFragment = CardDetailsFragment.from(
+                                    dropInRequest, cardNumber, configuration, hasTokenizationKeyAuth);
+                            replaceExistingFragment(cardDetailsFragment, CARD_DETAILS_TAG);
+                        } else {
+                            finishDropInWithError(authorizationError);
```
sarahkoop, line 349:
```suggestion
                            finishDropInWithError(configurationError);
```
I think this is for handling the config error, but we should also handle the auth error

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: This one is kind of covered actually. If authorization is `null` then `instanceof ClientToken` will return false.

Drop-In/src/main/java/com/braintreepayments/api/DropInActivity.java
sshropshire: Ah yeah good call. I'll update.

Drop-In/src/main/java/com/braintreepayments/api/DropInClient.java
```diff
@@ -66,7 +84,19 @@ public DropInClient(Context context, String authorization, DropInRequest dropInR
     }
 
     DropInClient(Context context, String authorization, String sessionId, DropInRequest dropInRequest) {
-        this(createDefaultParams(context, authorization, dropInRequest, sessionId));
+        this(createDefaultParams(context, authorization, null, dropInRequest, sessionId, null, null));
+    }
+
+    public DropInClient(FragmentActivity activity, DropInRequest dropInRequest, ClientTokenProvider clientTokenProvider) {
```
sarahkoop, line 90: I see we don't have doc strings on the original public constructor, but might be helpful to add them for all of these to clarify when to use each.

sshropshire: Yeah for sure maybe in a separate PR before release? That might be a good task to do in tandem with the docs site PR.

Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
```diff
@@ -94,46 +96,52 @@ void getPaymentMethodNonces(GetPaymentMethodNoncesCallback callback) {
...
 
     void deletePaymentMethod(final Context context, final PaymentMethodNonce paymentMethodNonce, final DeletePaymentMethodNonceCallback callback) {
-        boolean usesClientToken = braintreeClient.getAuthorization() instanceof ClientToken;
+        braintreeClient.getAuthorization(new AuthorizationCallback() {
+            @Override
+            public void onAuthorizationResult(@Nullable Authorization authorization, @Nullable Exception error) {
+                boolean usesClientToken = authorization instanceof ClientToken;
+
+                if (!usesClientToken) {
+                    Exception clientTokenRequiredError =
```
sarahkoop, line 105: This technically also handles when auth is null and there's an error, but do we want to return the explicit error in that case instead?

sshropshire: This actually calls back. The other one was guarding a UI update for vaulted payment methods.

sarahkoop (approved):
🔥 🔥

andrespelaezp (approved):
Looking nice!

PR #344, opened 2022-06-01
Add GitHub Actions Release Process

sarahkoop:
### Summary of changes

 - Add GitHub Actions release process

**Note:** We decided not to remove any of the old release process from the Rakefile, so that we can have a fallback if anything goes wrong with the GitHub Actions release. The Rakefile should be cleaned up eventually.

**Note:** This release process is only for current major version releases. The version number updating will not work with beta releases. TODO before the next major version beta release - add a beta release process to GitHub Actions.

 ### Checklist

- @mattwylder 
- @sarahkoop

### DIFF
--- a/.github/actions/decode_signing_key_action/action.yml
+++ b/.github/actions/decode_signing_key_action/action.yml
@@ -1,0 +1,18 @@
+name: Decode signing key
+description: 'Decodes gpg key into file'
+inputs:
+  signing_key_file:
+    description: 'Signing key file'
+    required: true
+    default: ''
+  signing_file_path:
+    description: 'Signing file path'
+    required: true
+    default: ''
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        echo "${{inputs.signing_key_file}}" > ~/secretKey.gpg.b64
+        base64 -d ~/secretKey.gpg.b64 > ${{ inputs.signing_file_path }}
+      shell: bash
--- a/.github/actions/publish_module/action.yml
+++ b/.github/actions/publish_module/action.yml
@@ -36,11 +36,11 @@ runs:
         SIGNING_KEY_ID: ${{ inputs.signing_key_id }}
         SIGNING_KEY_PASSWORD: ${{ inputs.signing_key_pwd }}
         SIGNING_KEY_FILE: ${{ inputs.signing_key_file }}
-#    - run: ./gradlew --stacktrace closeAndReleaseRepository
-#      shell: bash
-#      env:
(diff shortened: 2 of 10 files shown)


.github/actions/publish_module/action.yml
```diff
@@ -0,0 +1,46 @@
...
+  steps:
+    - run: ./gradlew --stacktrace clean :${{ inputs.module }}:publishToSonatype
+      shell: bash
+      env:
+        SONATYPE_USERNAME: ${{ inputs.sonatype_usr }}
+        SONATYPE_PASSWORD: ${{ inputs.sonatype_pwd }}
+        SIGNING_KEY_ID: ${{ inputs.signing_key_id }}
+        SIGNING_KEY_PASSWORD: ${{ inputs.signing_key_pwd }}
+        SIGNING_KEY_FILE: ${{ inputs.signing_key_file }}
+    - run: ./gradlew --stacktrace closeAndReleaseRepository
```
sarahkoop, line 39: This step hasn't been tested - we tested up to `publishToSonatype` and manually dropped the repos from Sonatype without releasing. The `closeAndReleaseRepository` step should actually release the SDK, so we did not want to do that in testing.

.github/actions/set_github_user/action.yml
```diff
@@ -0,0 +1,9 @@
+name: Set GitHub User
+description: 'Sets github user for a particular workflow'
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        git config user.name paypalsdk
```
sarahkoop, line 7: We took this user from the Unified SDK - is this correct or do we want to still use braintreeps?

.github/actions/decode_signing_key_action/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    description: 'Signing file path'
+    required: true
+    default: ''
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        echo "${{inputs.signing_key_file}}" > ~/secretKey.gpg.b64
+        base64 -d ~/secretKey.gpg.b64 > ${{ inputs.signing_file_path }}
+      shell: bash
```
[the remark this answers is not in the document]
mattwylder: Seems like Github Actions expects an object with two keys, `run` and `shell`. The syntax highlighting on Github is messed up. It makes it look like `shell` is part of the multiline string, but they're two keys in the same object.

jaxdesmarais (approved):
⛵

sshropshire (approved):

PR #347, opened 2022-06-06
Fix issue where UserCanceledException is not being sent

### LINKED ISSUE #346: No result on user dismiss
### General information

* SDK/Library version: 6.1.0
* Environment: sandbox
* Android Version and Device: any

### Issue description
On version 6.1.0 I don't receive any result when the user dismisses the drop in.
I think the intended behaviour was to return a `DropInResult` with `UserCanceledException` from `DropInListener#onDropInFailure`. However, in `DropInActivityResultContract#parseResult` this logic is conditioned to `
intent != null`, which is always false for `RESULT_CANCELED`, leading to no result being produced.

sshropshire:
### Summary of changes

 - Fixes #346
 - This PR fixes an issue where `UserCanceledException` was not being sent properly.

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Braintree Android Drop-In Release Notes
 
+## unreleased
+
+* Fix issue where `UserCanceledException` was not being sent on bottom sheet dismissal.
+
 ## 6.1.0
+
 * Update Vault Manager inline documentation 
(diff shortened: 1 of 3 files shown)


jaxdesmarais (approved):
⛵

andrespelaezp (approved):

PR #401, opened 2023-02-27
Add Browser Switch Note in Migration Guide

sshropshire:
### Summary of changes

 - Add note in Migration Guide that Browser Switch is configured automatically in DropIn
 - Fix for [braintree/braintree_android #692](https://github.com/braintree/braintree_android/issues/692)

 ### Checklist

### Authors

- @sshropshire

### DIFF
--- a/v6_MIGRATION_GUIDE.md
+++ b/v6_MIGRATION_GUIDE.md
@@ -14,6 +14,7 @@ v6 of the Drop-In SDK requires v4 of the [Braintree Android SDK](https://github.
 1. [Launch Drop-In](#launch-drop-in)
 1. [Handle Drop-In Result](#handle-drop-in-result)
 1. [Fetch Last Used Payment Method](#fetch-last-used-payment-method)
+1. [Browser Switch](#browser-switch)
 
 ## Gradle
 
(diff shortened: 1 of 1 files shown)


jaxdesmarais (approved):

v6_MIGRATION_GUIDE.md
```diff
@@ -244,3 +245,8 @@ Kotlin:
         // handle result
     }
 ```
+
+## Browser Switch
+
+DropIn handles browser switching internally. Any payment method that requires a browser switch will work automatically.
```
scannillo, line 251: Do they no longer need to include an intent-filter in their AndroidManifest.xml? If so, I think we should call that out specifically so they know what to remove.

sshropshire: Yeah good call, i'll update. This slipped through the cracks I don't think an intent filter was ever needed 😅 .

scannillo (approved):

v6_MIGRATION_GUIDE.md
sshropshire: Definitely yeah thanks for finding this. We did it for 3DS but I missed that one, I'll update.

PR #409, opened 2023-03-31
Only Allow Google Pay Launch When an Activity is in the Foreground

sshropshire:
### Summary of changes

 - To prevent a double tap from launching Google Pay twice, make sure the Activity is in the foreground before presenting the Google Pay flow

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


jaxdesmarais (approved):

scannillo (approved):
🦸‍♂️

PR #438, opened 2023-09-18
Upgrade to Android 14

sarahkoop:
### Summary of changes

 - Update compile and target SDK version to API 31
 - Bump braintree_android to 4.38.1

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## unreleased
 
 * Bump braintree_android module dependency versions to `4.38.1`
-* Upgrade `compileSdkVersion` and `targetSdkVersion` to API 34
+* Android 14 Support
+  * Upgrade `compileSdkVersion` and `targetSdkVersion` to API 34
 
 ## 6.12.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## unreleased
 
-* Bump braintree_android module dependency versions to `4.38.0`
+* Bump braintree_android module dependency versions to `4.38.1`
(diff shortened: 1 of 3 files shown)


Drop-In/build.gradle
```diff
@@ -47,7 +47,9 @@ android {
         warning 'GradleCompatible'
         textReport true
         textOutput 'stdout'
-        disable 'InvalidPackage' // The Cardinal SDK is referencing the `javax.naming` package but it is unavailable
+        // RestrictedApi - Accessing Braintree Core SDK non-public methods
+        // InvalidPackage - The Cardinal SDK is referencing the `javax.naming` package but it is unavailable
+        disable 'RestrictedApi', 'InvalidPackage'
```
sshropshire, line 52: random: I wonder if Cardinal actually needs `javax.naming` to work

sarahkoop: Actually looks like they don't use it anymore or it's not causing a lint error - removed that option 👍

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## unreleased
 
-* Bump braintree_android module dependency versions to `4.38.0`
+* Bump braintree_android module dependency versions to `4.38.1`
+* Upgrade `compileSdkVersion` and `targetSdkVersion` to API 34
```
jaxdesmarais, line 6:
```suggestion
* Android 14 Support
  * Upgrade `compileSdkVersion` and `targetSdkVersion` to API 34
```
^ we got some merchant feedback about adding a note for Android 14 explicitly on the BT side so I YOLO pushed a commit there, but we should do the same here.

jaxdesmarais (approved):

PR #440, opened 2023-09-19
Migration Guide Custom URL Scheme

sshropshire:
### Summary of changes

 - Add migration guide entry to explain custom URL scheme override through `DropInRequest`

 ### Checklist

- @sshropshire

### DIFF
--- a/v6_MIGRATION_GUIDE.md
+++ b/v6_MIGRATION_GUIDE.md
@@ -256,6 +256,7 @@ Internally we use a manifest placeholder to support deep linking into `DropInAct
 
 ```xml
 <activity android:name="com.braintreepayments.api.DropInActivity"
+  android:exported="true"
   tools:node="merge"
   >
   <intent-filter>
--- a/v6_MIGRATION_GUIDE.md
+++ b/v6_MIGRATION_GUIDE.md
@@ -259,6 +259,7 @@ Internally we use a manifest placeholder to support deep linking into `DropInAct
   android:exported="true"
   tools:node="merge"
   >
+  <intent-filter tools:node="removeAll" />
   <intent-filter>
     <action android:name="android.intent.action.VIEW" />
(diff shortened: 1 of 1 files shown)


v6_MIGRATION_GUIDE.md
```diff
@@ -250,3 +250,30 @@ Kotlin:
 
 DropIn handles browser switching internally. Specifying an `<intent-filter />` in `AndroidManifest.xml` is no longer required. Any payment method that requires a browser switch will work automatically.
 
+On the other hand, there are some scenarios that will require you to set a custom URL scheme for browser switch deep linking. For example, if your `applicationId` contains uppercase letters, you will need to override the default deep link configuration set by the DropIn library.
+
+Internally we use a manifest placeholder to support deep linking into `DropInActivity`. You can override the deep link intent filter for `DropInActivity` by placing the following xml snippet in your app's `AndroidManifest.xml`:
```
sarahkoop, line 255: Should we also add a line like: If you are not setting a custom URL scheme for deep linking, you do not need to add an intent filter to support Drop-in?

In case merchants see this an think they always need to add this to their manifest.

sshropshire: Yeah I thought about adding another section for this to make it clear. Technically we say "no additional configuration required" in the first paragraph of this section, but it may not be prominent enough.

sarahkoop: Oh true ok that's probably enough

jaxdesmarais (approved):

v6_MIGRATION_GUIDE.md
scannillo: Do we have a ticket for adding this to the public dev docs?

sshropshire: Ended up creating DTBTSDK-3020 for this.

PR #468, opened 2024-03-27
[v7 Beta] Create DropInLauncher

sshropshire:
### Summary of changes

 - Replace `DropInClient` with `DropInLauncher`
 - Add `DropInLauncherCallback` to receive Activity result
 - Move `fetchMostRecentPaymentMethod()` to `RecentPaymentMethodsClient`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
@@ -188,7 +188,7 @@ public class MainActivity extends BaseActivity implements DropInListener {
         }
 
         // TODO: make auth string a required DropInRequest constructor parameter
-        dropInLauncher.launchDropIn(authString, dropInRequest);
+        dropInLauncher.start(authString, dropInRequest);
     }
 
     private ThreeDSecureRequest demoThreeDSecureRequest() {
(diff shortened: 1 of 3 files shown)


Demo/src/main/java/com/braintreepayments/demo/MainActivity.java
```diff
@@ -152,7 +187,8 @@ public void launchDropIn(View v) {
             dropInRequest.setThreeDSecureRequest(demoThreeDSecureRequest());
         }
 
-        dropInClient.launchDropIn(dropInRequest);
+        // TODO: make auth string a required DropInRequest constructor parameter
+        dropInLauncher.launchDropIn(authString, dropInRequest);
```
jaxdesmarais, line 191: This reads a little odd to me since it's like the inverse of one another, not sure if something like this would make it feel less awward:
```suggestion
        dropInLauncher.start(authString, dropInRequest);
```

sshropshire: Yeah def that's a good call. It aligns nicely with our SDKs as a whole. I'll do this in a separate commit to make sure all occurrences are renamed.

PR #472, opened 2024-04-17
Pass hasUserLocationConsent to DataCollector

tdchow:
### Summary of changes

 - Add javadocs for `hasUserLocationConsent` param
 - Pass `hasUserLocationConsent` to `DataCollector.collectDeviceData()`

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/DropInInternalClient.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/DropInInternalClient.java
@@ -95,8 +95,8 @@ class DropInInternalClient {
         boolean hasUserLocationConsent,
         DataCollectorCallback callback
     ) {
-        // TODO: pass in hasUserLocationConsent
-        dataCollector.collectDeviceData(activity, callback);
+        DataCollectorRequest request = new DataCollectorRequest(hasUserLocationConsent);
+        dataCollector.collectDeviceData(activity, request, callback);
     }
 
     void performThreeDSecureVerification(final FragmentActivity activity, PaymentMethodNonce paymentMethodNonce, final DropInResultCallback callback) {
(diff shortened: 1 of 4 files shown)


Drop-In/src/main/java/com/braintreepayments/api/DropInInternalClient.java
```diff
@@ -148,10 +148,8 @@ void shouldRequestThreeDSecureVerification(PaymentMethodNonce paymentMethodNonce
     void tokenizePayPalRequest(FragmentActivity activity, PayPalFlowStartedCallback callback) {
         PayPalRequest paypalRequest = dropInRequest.getPayPalRequest();
         if (paypalRequest == null) {
-            paypalRequest = new PayPalVaultRequest();
+            paypalRequest = new PayPalVaultRequest(false);
         }
-        // TODO: set hasUserLocationConsent on paypalRequest
-        dropInRequest.hasUserLocationConsent();
```
tdchow, line 154: `hasUserLocationConsent` is set on the `PayPalRequest` directly. No need to set it here.

sarahkoop: If the merchant has set `hasUserLocationConsent` on their `dropInRequest` already - can we pass that value in when creating the default `PayPalVaultRequest`?

tdchow: yep, great suggestion! i'll update this.

unrelated question - if this function gets called with a null `PayPalRequest` and a default one is created, are there scenarios where `payPalClient.tokenizePayPalAccount()` will ever succeed? are there required fields on the request?

sarahkoop: Before this change, there were no required fields for a `PayPalVaultRequest` (checkout request requires amount). So if a merchant doesn't pass a custom `PayPalRequest` to their `DropInRequest`, drop-ins default behavior is the PayPal billing agreement/vault flow - it should succeed when we create the default vault request.

sarahkoop:
Is there going to be a separate PR to pass this flag to all other instances of collectDeviceData throughout drop-in?

tdchow:
> Is there going to be a separate PR to pass this flag to all other instances of collectDeviceData throughout drop-in?

I was intending this to be the final PR for drop-in. I wasn't able to find any other instances where we'd need to pass the flag when searching for "collectDeviceData" in the repo. Are there other callsites that I'm currently missing?

sarahkoop (approved):

saperi22 (approved):

PR #497, opened 2025-04-18
update braintree_android dep

saperi22:
### Summary of changes

 - Update braintree_android to 4.50.0. Adds support for Android 15.

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
+++ b/Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
@@ -145,7 +145,7 @@ public class PaymentMethodClient {
                     input.put(SINGLE_USE_TOKEN_ID, paymentMethodNonce.getString());
                     variables.put(INPUT, input);
                     base.put(VARIABLES, variables);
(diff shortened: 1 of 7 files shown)


Drop-In/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
```diff
@@ -145,7 +145,7 @@ public void onAuthorizationResult(@Nullable Authorization authorization, @Nullab
                     input.put(SINGLE_USE_TOKEN_ID, paymentMethodNonce.getString());
                     variables.put(INPUT, input);
                     base.put(VARIABLES, variables);
-                    base.put(GraphQLConstants.Keys.OPERATION_NAME,
```
saperi22, line 148: It looks like `GraphQLConstants.Keys.OPERATION_NAME` constant was removed from the braintree_android sdk that was missed during 4.50.0 publish. As it is only a constant, it is being hardcoded here.

tdchow (approved):

sarahkoop (approved):

jaxdesmarais (approved):

PR #510, opened 2026-03-09
https://github.com/braintree/braintree_android/pull/1538

jaxdesmarais:
### Summary of changes

 - Update for security fix to validate SDK version via regex to prevent malicious code from running

 ### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,35 @@ on:
 env:
   SIGNING_KEY_FILE_PATH: /home/runner/secretKey.gpg
 jobs:
+  validate_version:
+    name: Validate Version Input
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate version input
+        env:
+          INPUT_VERSION: ${{ github.event.inputs.version }}
+        run: |
+          set -euo pipefail
+
+          # Validate version format: x.x.x or x.x.x-betax (e.g., 4.0.0, 4.0.0-beta1)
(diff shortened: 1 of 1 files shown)


saperi22 (approved):

.github/workflows/release.yml
```diff
@@ -84,6 +113,7 @@ jobs:
         uses: ./.github/actions/set_github_user
       - name: Update Version
         run: |
+          set -euo pipefail
```
saralvasquez, line 116: For my knowledge, what does this do?

saralvasquez (approved):
This makes sense to me, but where did this come from? Was there an version input issue I missed?

.github/workflows/release.yml
jaxdesmarais: That's a great question! It makes it so that if the update version validation fails the release process exits/fails fast vs the error getting silently swallowed and continuing on. Each portion of the tag means:
- `-e`= Exit immediately if any command returns a non-zero exit code            
- `-u` = Treat unset variables/typos as errors
- `-o pipefail` = If any command in a pipeline fails, the whole pipeline fails

jaxdesmarais:
> This makes sense to me, but where did this come from? Was there an version input issue I missed?

@saralvasquez yes, there were a number of security inbounds that came to me internally. To allow the team to focus on other higher priority work, I've gone ahead and just handled tackling them here and there as I had time.

